PR #12, opened 2019-12-12
fix: LAST_INSERT_ID() breaks on some MySQL compliant databases

cyjake:
even with `UPDATE id = LAST_INSERT_ID(id)` hack https://stackoverflow.com/questions/778534/mysql-on-duplicate-key-last-insert-id

cc @chenhui5416


cyjake:
v0.4.5 released

PR #43, opened 2020-04-10
build: enable mysql service

cyjake:
att


cyjake:
> On Unix, MySQL programs treat the host name localhost specially, in a way that is likely different from what you expect compared to other network-based programs: the client connects using a Unix socket file.

Lesson learned the hard (time wasting) way.

Issue #46, opened 2020-11-13
Model.build(values, options)


cyjake:
options are not handled correctly

PR #47, opened 2020-11-17
feat: enhance sequelize adapter & .nodeify() support

cyjake:
enhanced following static methods on Model:

- Model.build()
- Model.count()
- Model.min()
- Model.max()
- Model.sum()

also added `.finally()` and `.nodeify()` for better integration with project at hand


cyjake:
fixes #46

PR #58, opened 2020-12-10
release: v1.2.0

cyjake:
att

### DIFF
--- a/lib/bone.js
+++ b/lib/bone.js
@@ -916,7 +916,7 @@ class Bone {
         connection.release();
       }
     } else if (callback.constructor.name === 'GeneratorFunction') {
-      const gen = callback();
+      const gen = callback({ connection });
       let result;
 
       try {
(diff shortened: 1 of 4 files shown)


test/unit/test.realm.js
```diff
@@ -101,21 +101,10 @@ describe('=> Realm', () => {
...
-        users (gmt_create, email, nickname)
-        VALUES (?, ?, ?)`,[
-          new Date(),
-          email,
-          'Loki',
-        ]);
+      await realm.transaction(function *({ connection }) {
+        const sql = 'INSERT INTO users (gmt_create, email, nickname) VALUES (?, ?, ?)';
+        yield realm.query(sql, [ new Date(), email, 'Thor' ], { connection });
+        yield realm.query(sql, [ new Date(), email, 'Loki' ], { connection });
```
cyjake, line 107: currently, automatic connection injection in generator function is only possible if Spell were involved. Hence if it's raw query, we need to pass on `{ connection }` manually.

lib/bone.js
```diff
@@ -916,7 +916,7 @@ class Bone {
         connection.release();
       }
     } else if (callback.constructor.name === 'GeneratorFunction') {
-      const gen = callback();
+      const gen = callback({ connection });
```
cyjake, line 919: make sure generator function gets connection as well

JimmyDaddy (approved):

PR #61, opened 2020-12-15
fix: set default values

JimmyDaddy:
set defaultValue when create or bulkCreate

### DIFF
--- a/lib/bone.js
+++ b/lib/bone.js
@@ -365,8 +365,9 @@ class Bone {
       const value = this.attribute(name);
       if (value != null) data[name] = value;
       // set defaultValues
-      if (!value && !attributes[name].allowNull && attributes[name].defaultValue !== undefined && attributes[name].defaultValue !== null) {
-        this[name] = attributes[name].defaultValue;
+      const { defaultValue } = attributes[name];
+      if (value === undefined && defaultValue != null) {
+        this[name] = defaultValue;
       }
     }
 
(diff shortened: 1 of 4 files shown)


lib/bone.js
```diff
@@ -757,14 +761,32 @@ class Bone {
...
+     */
+    const _setDefaultValue = (record) => {
+      for (const name in attributes) {
+        const value = record[name];
+        // set defaultValues
+        if (!value && !attributes[name].allowNull
+          && attributes[name].defaultValue !== undefined
+          && attributes[name].defaultValue !== null) {
+          record[name] = attributes[name].defaultValue;
+        }
```
cyjake, line 778:
```suggestion
const { defaultValue }  = attributes[name];
if (value === undefined && defaultValue != null) {
  record[name] = defaultValue
}
```

value might be falsy values such as 0 or '', in which case the defaultValue should not take precedence. It is also possible that the value is null and defaultValue is something else (such as 0), in that case value = null should not be overwritten.

allowNull should not take part here. if defaultValue is declared, it should be used whenever possible.

lib/bone.js
```diff
@@ -757,14 +762,31 @@ class Bone {
...
+    const { driver, table, attributes, primaryKey, primaryColumn } = this;
+    /**
+     * set defaultValue when create
+     * @param {object} values
+     * @memberof Bone [private]
+     */
+    const _setDefaultValue = (record) => {
+      for (const name in attributes) {
+        const value = record[name];
+        // set defaultValues
```
cyjake, line 774: unnecessary comment

lib/bone.js
```diff
@@ -364,6 +364,11 @@ class Bone {
     for (const name in attributes) {
       const value = this.attribute(name);
       if (value != null) data[name] = value;
+      // set defaultValues
+      const { defaultValue } = attributes[name];
+      if (value === undefined && defaultValue != null) {
+        this[name] = defaultValue;
```
cyjake, line 370: should be `data[name] = defaultValue` here

L368-L371 can be wrapped in the else branch of `if (value != value)` or vice versa

cyjake (approved):

PR #65, opened 2020-12-22
fix: init with custom setters

JimmyDaddy:
fix: init with custom setters


lib/bone.js
```diff
@@ -760,14 +759,14 @@ class Bone {
...
-    const _setDefaultValue = (record) => {
+    const _setDefaultValue = record => {
       for (const name in attributes) {
         const value = record[name];
         // set defaultValues
         const { defaultValue } = attributes[name];
         if (value === undefined && defaultValue != null) record[name] = defaultValue;
       }
-    }
+    };
```
cyjake, line 769: regular function should be preferred if arrow function is not necessary, since there's no `this` to bind

lib/bone.js
```diff
@@ -948,7 +947,13 @@ class Bone {
       };
     }
 
-    Object.defineProperties(this.prototype, descriptors);
+    const customDescriptors = {};
+    Object.keys(descriptors).map(key => {
+      const descriptor = Object.getOwnPropertyDescriptor(descriptors, key);
+      customDescriptors[key] = descriptor;
+    });
```
cyjake, line 954: You can grab all the descriptors using [Object.getOwnPropertyDescripts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors) directly.

I think we can call the third argument `overrides`, e.g. `init(attributes, opts, overrides)`, and then `const descriptors = Object.getOwnPropertyDescripts(overrides)` here.

btw, as we discussed, is `configurable: true` still necessary here?

test/integration/suite/basics.test.js
```diff
@@ -182,6 +183,28 @@ describe('=> Accessors', function() {
     });
     expect(user.isValid).to.eql(false);
   })
+
+  it('Bone.init(arrt, opts, descriptors) should work with setter', async () => {
```
cyjake, line 187: attrs

test/models/user.js
```diff
@@ -2,11 +2,28 @@
...
     return this.status === 1;
   }
-};
+}
+
+Object.defineProperty(User, 'formatter', {
+  get() {
+    return formatter;
+  },
+});
```
cyjake, line 26: maybe `static get formatter() {}` should be preferred?

test/models/user.js
```diff
@@ -30,11 +47,17 @@ User.init({
...
-    get() {
-      return this.status !== 1;
+  get isValid() {
+    return this.status !== 1;
+  },
+  set nickname(value) {
+    if (value === 'Zeus') {
+      this.attribute('nickname', 'V');
+    } else {
+      const { formatter: format } = this.constructor;
```
cyjake, line 57: format is a verb, which makes `format.formatName` a bit weird to read. Hence I'd prefer to leave it as `formatter` here.

cyjake (approved):
+1

PR #66, opened 2020-12-23
[AOP]Feat/hooks

JimmyDaddy:
### hooks

> before/after

- create
   - instance
      - instance.beforeCreate(instance, ...args)
      - instance.afterCreate(instance, createResult)
   - class(static) Proxy to instance
- update
   - instance
      - beforeUpdate(instance, ...args)
      - afterUpdate(instance, updateResult)
   - class
      - beforeUpdate(...args)
      - afterUpdate(updateResult, class)
- remove
   - Same as update
- upsert
   - instance
      - beforeUpsert(instance, ..args)
      - afterUpsert(updateResult, ...args)
   - class(sequelize mode only)
      - beforeUpsert(..args)
      - afterUpsert(updateResult, class)
- destroy (sequelize mode)
   - instance
      - beforeDestory(instance, ...args)
      - afterDestory(instance, destoryResult)
   - class
      - beforeDestory(..args)
      - afterDestory(destoryResult, class)
- save
   > save should trigger create|update|upsert hooks
   - instance
       - beforeSave(instance, ...args)
       - afterSave(instance, result)

class.addHook(hookName, fnNameOrFun, fun)
### removeHook

not implemented

Related to #71


test/unit/validator.test.js
```diff
@@ -78,26 +78,26 @@ describe('validator', () => {
     });
   });
 
-  afterEach(async () => {
```
cyjake, line 81: `async () => {}` with a space after `async` should be preferred. Was there a rule change in eslintrc?

lib/setup_hooks.js
```diff
@@ -0,0 +1,165 @@
+'use strict';
+
+/**
+ *
+ * @param {*} isInstance class or instance
+ * @param {*} fnName
+ * @param {arguments} args
+ * @param {*} target class or instance
```
cyjake, line 8:
```suggestion
 * @param {boolean} isInstance class or instance
 * @param {string} fnName
 * @param {Array} args
 * @param {Bone} target class or instance
```

test/unit/validator.test.js
JimmyDaddy: I added an eslint rule `space-before-function-paren` to avoid `funcName   (){}` , I'll remove it

cyjake (approved):

PR #68, opened 2020-12-24
fix: sequelize mode isNewRecord

JimmyDaddy:
rt

### DIFF
--- a/lib/bone.js
+++ b/lib/bone.js
@@ -84,6 +84,13 @@ class Bone {
       rawUnset: new Set(),
     }));
 
+    Object.defineProperty(this, 'isNewRecord', {
+      value: true,
+      enumerable: false,
+      writable: true,
+      configurable: true,
+    });
+
     if (dataValues) {
       for (const name in dataValues) {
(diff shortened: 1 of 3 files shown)


lib/bone.js
```diff
@@ -83,6 +83,14 @@ class Bone {
       rawInitial: {},
       rawUnset: new Set(),
     }));
+
+    Object.defineProperty(this, 'isNewRecord', {
+      value: true,
+      enumerable: false,
+      writable: true,
+      configurable: true,
+    });
```
cyjake, line 92: A direct assignment, such as `this.isNewRecord = true`, shall suffice.

cyjake: btw, I did take a look in Active Record of Ruby on Rails. The equivalent property is called `new_record?` (notice the `?` at the end, which indicates it's a boolean). Hence I think `isNewRecord` is perfect for us.

JimmyDaddy: `isNewRecord` should not be  enumerable, a direct assignment will make developer confuse

cyjake: good point.

PR #78, opened 2021-01-17
refactor: replace attributeChanged() with changed()

cyjake:
... which seems a little refactor on changed(name) is necessary

- https://www.yuque.com/leoric/guide/dirty

### DIFF
--- a/lib/bone.js
+++ b/lib/bone.js
@@ -215,8 +215,8 @@ class Bone {
   }
 
   /**
-   * @deprecated {attributeChanged} will be deprecated, RECOMMEND to use {@link Bone.changed}
-   * Check if the value of attribute is changed or not ({@link Bone.rawSaved}).
+   * @deprecated {attributeChanged} is deprected, use {@link Bone#changed} instead
+   * Check if the value of attribute is changed or not ({@link Bone#rawSaved}).
    * @param {string} name - attribute name
    * @example
    * const post = await Post.findOne({ title: 'Leah' })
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -42,14 +42,14 @@
...
+    "dayjs": "^1.10.3",
+    "eslint": "^7.18.0",
     "expect.js": "^0.3.1",
     "jsdoc": "^3.6.3",
-    "mocha": "^5.2.0",
+    "mocha": "^8.2.1",
     "mysql": "^2.17.1",
     "mysql2": "^1.7.0",
-    "nyc": "^13.1.0",
+    "nyc": "^5.6.0",
```
cyjake, line 52: upgraded several dev dependencies

lib/bone.js
```diff
@@ -281,18 +281,20 @@ class Bone {
    */
   changes(name) {
     if (name != null) {
-      if (this.rawUnset.has(name) || this.rawSaved[name] === undefined) return {};
+      if (this.rawUnset.has(name)) return {};
```
cyjake, line 284: If value is not changed before, consider the attribute is changed as well

lib/drivers/abstract/index.js
```diff
@@ -40,13 +40,25 @@ module.exports = class AbstractDriver {
...
+      if (type === JSON && typeof value.toObject === 'function') {
+        return JSON.stringify(value.toObject());
+      }
+      if (type === Date && typeof value.toDate === 'function') {
+        return value.toDate();
+      }
+      if (type === String && typeof value.toString === 'function') {
+        return value.toString();
+      }
+    }
```
cyjake, line 60: Try calling object.toXxx() methods to cast values in a customized way

JimmyDaddy (approved):

PR #81, opened 2021-01-19
fix:  #80 [sequelize] getterMethods should be executed while calling toJSON() or toObject()

JimmyDaddy:


lib/bone.js
```diff
@@ -326,9 +326,16 @@ class Bone {
...
     const obj = {};
 
-    for (const name in this.constructor.attributes) {
-      if (!this.rawUnset.has(name) && this[name] != null) {
-        obj[name] = this[name];
+    const proto = Object.getPrototypeOf(this);
+    const propertyNames = Object.getOwnPropertyNames(proto);
+    const attributeNames = Object.keys(this.constructor.attributes);
+    for (const key of propertyNames) {
+      if (attributeNames.includes(key) && this.rawUnset.has(key)) continue;
```
cyjake, line 333:
```suggestion
      if (this.rawUnset.has(key)) continue;
```

`this.rawUnset` is a subset of attribute keys, hence there's no need to check against attribute names before rawUnset here.

lib/bone.js
```diff
@@ -352,8 +359,18 @@ class Bone {
...
     const obj = {};
 
-    for (const name in this.constructor.attributes) {
-      if (!this.rawUnset.has(name)) obj[name] = this.attribute(name);
+    const proto = Object.getPrototypeOf(this);
+    const propertyNames = Object.getOwnPropertyNames(proto);
+    const attributeNames = Object.keys(this.constructor.attributes);
+
+    for (const key of propertyNames) {
+      if (attributeNames.includes(key) && this.rawUnset.has(key)) continue;
```
cyjake, line 367:
```suggestion
      if (this.rawUnset.has(key)) continue;
```

PR #82, opened 2021-01-19
fix: instance[key] should not change after updating failed

JimmyDaddy:
fix: instance[key] should not change after updating failed
eg.
```js
const user = new User({
    email: 'a@e.com',
    nickname: 'sss',
    fingerprint: 'finger111'
});
try {
    // assume it will be failed
    await user.update({ fingerprint: 'aaa' });
} catch (err) {}

user.fingerprint // should be 'finger111', not 'aaa'


```

### DIFF
--- a/lib/bone.js
+++ b/lib/bone.js
@@ -305,7 +305,7 @@ class Bone {
     }
     const result = {};
     for (const attrKey of Object.keys(this.constructor.attributes)) {
-      if (this.rawUnset.has(name)) continue;
+      if (this.rawUnset.has(attrKey)) continue;
       const value = this.attribute(attrKey);
       const valueWas = this.attributeWas(attrKey);
 
@@ -492,12 +492,12 @@ class Bone {
       }
     } else {
       for (const name in opts) {
-        const originValue = this[name];
+        const originValue = this.attribute(name);
         // exec custom setters in case it exist
         this[name] = opts[name];
         changes[name] = this.attribute(name);
         // revert value in case update failed
-        this[name] = originValue;
+        this.attribute(name, originValue);
       }
     }
 
@@ -518,9 +518,9 @@ class Bone {
     }
     const spell = new Spell(Model).$where(where).$update(changes);
     return spell.later(result => {
-      // sync changes
+      // sync changes (changes has been formatted by custom setters, use this.attribute(name, value) directly)
       for (const key in changes) {
-        this[key] = changes[key];
+        this.attribute(key, changes[key]);
       }
       this.syncRaw(changes);
       return result.affectedRows;
--- a/lib/bone.js
+++ b/lib/bone.js
@@ -164,7 +164,7 @@ class Bone {
     const changes = this.changes();
     let changedValue = {};
     for (const key in changes) {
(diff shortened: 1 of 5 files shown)


lib/bone.js
```diff
@@ -160,10 +160,26 @@ class Bone {
    */
   _validateAttributes(values = {}) {
     const { attributes } = this.constructor;
-    Object.keys(this.raw).map((valueKey) => {
+    // check changed values
+    const changes = this.changes();
+    let changedValue = {};
+    for (const key in changes) {
+      if (changes[key].length && changes[key].length === 2) {
```
cyjake, line 167:
```suggestion
      if (changes[key].length === 2) {
```

`changes[key].length &&` seems unnecessary

lib/bone.js
```diff
@@ -160,10 +160,26 @@ class Bone {
    */
   _validateAttributes(values = {}) {
     const { attributes } = this.constructor;
-    Object.keys(this.raw).map((valueKey) => {
+    // check changed values
+    const changes = this.changes();
+    let changedValue = {};
```
cyjake, line 165: changedValues

lib/bone.js
```diff
@@ -160,10 +160,26 @@ class Bone {
...
+      if (changes[key].length && changes[key].length === 2) {
+        changedValue[key] = changes[key][1];
+      }
+    }
+
+    // merge all changed values
+    changedValue = {
+      ...changedValue,
+      ...values,
+    };
```
cyjake, line 176: Personally I'd prefer `Object.assign(changedValues, values)`

lib/bone.js
```diff
@@ -160,10 +160,26 @@ class Bone {
...
+
+    Object.keys(changedValue).map((valueKey) => {
       const attribute = attributes[valueKey];
+      if (!attribute) return;
```
cyjake, line 180:
```suggestion
    for (const valueKey of Object.keys(changedValue)) {
      const attribute = attributes[valueKey];
      if (attribute == null) return;
```

lib/validator.js
```diff
@@ -49,6 +49,9 @@ const validators = {
   is(str, pattern, modifiers) {
     return this.regex(str, pattern, modifiers);
   },
+  notEmpty(str) {
+    return !str.match(/^[\s\t\r\n]*$/);
```
cyjake, line 53: no need to match the complete string, test for non whitespace character shall suffice

```suggestion
    return /[^\s]/.test(str);
```

cyjake: btw, `\s` includes `[\t\r\n]`

test/integration/suite/basics.test.js
```diff
@@ -306,6 +307,13 @@ describe('=> Accessors', function() {
     expect(user.status).to.eql(1);
     expect(user.raw.status).to.equal(-1);
     await user.update({ status: 2 });
+    //set status(value = 0) {
```
cyjake, line 310:
```suggestion
    // set status(value = 0) {
```

cyjake (approved):
+1

PR #83, opened 2021-01-19
fix: disable attributes checking while init

JimmyDaddy:


test/unit/adapters/sequelize.test.js
```diff
@@ -868,6 +868,15 @@ describe('model.init with getterMethods and setterMethods', () => {
     assert.equal(user.fingerprint, 'Bloodborne');
     assert.equal(user.raw.fingerprint, encrypt('Bloodborne'));
   });
+
+  it('should work while init', async () => {
```
cyjake, line 872: 'should accept arbitrary properties'

lib/bone.js
```diff
@@ -186,7 +183,7 @@ class Bone {
     const { attributes } = this;
     Object.keys(values).map((valueKey) => {
       const attribute = attributes[valueKey];
-      if (!attribute) throw new Error(`${this.name} has no attribute "${valueKey}"`);
+      if (!attribute) return;
```
cyjake, line 186: a comment explaining this early return should be necessary

cyjake (approved):
+1

PR #84, opened 2021-02-08
fix: setter getter bug in multiple extends

JimmyDaddy:


lib/bone.js
```diff
@@ -47,6 +48,8 @@ function compare(attributes, columns) {
   return diff;
 }
 
+const privateKeys = ['raw', 'rawSaved', 'rawUnset', 'rawPrevious', 'isNewRecord'];
```
JimmyDaddy, line 51: 将这几个 私有 key 转为 Symbol 的方式处理会不会好一点

lib/utils/index.js
```diff
@@ -11,7 +11,27 @@ function compose() {
   return funcs.reverse().reduce((a, b) => (...arg) => b(a(...arg)));
 }
 
+function getPropertyNames (obj) {
+  if (obj == null) return [];
+  const propertyNames = [];
+  // avoid to deep clone obj
+  let muteObj = Object.getPrototypeOf(obj);
+  do {
+    propertyNames.push.apply(propertyNames, Object.getOwnPropertyNames(muteObj));
```
cyjake, line 20:
```suggestion
    propertyNames.push(...Object.getOwnPropertyNames(muteObj))
```

lib/utils/index.js
```diff
@@ -11,7 +11,27 @@ function compose() {
...
+    muteObj = Object.getPrototypeOf(muteObj);
+  } while (muteObj);
```
cyjake, line 22: We can stop early if `muteObj.constructor.name === 'Bone'` or `muteObj.constructor.attributes` exists

lib/utils/index.js
```diff
@@ -11,7 +11,27 @@ function compose() {
...
+  muteObj = {};
+  // get own properties
+  propertyNames.push.apply(propertyNames, Object.getOwnPropertyNames(obj));
+  // get unique properties' names, filter __proto__
+  for (const name of propertyNames) {
+    muteObj[name] = 1;
+  }
```
cyjake, line 29: How about use `Set()` in the first place? Could be handier than filtering with object key.

test/unit/adapters/sequelize.test.js
```diff
@@ -836,6 +837,12 @@ describe('model.init with getterMethods and setterMethods', () => {
     }
   });
 
+  // multiple implement
+  class UU extends User {
+    get j () {
+      return 'j';
+    }
+  }
```
cyjake, line 845: `class CustomUser` might be preferred. If this fixture is only used in the `toJSON()` test case, should've considered putting it within that test. Or better create another specific test describing intended test purpose.

lib/bone.js
```diff
@@ -91,18 +98,30 @@ class Bone {
...
+    }, {
+      key: rawPrevious,
+      value: {},
+    }, {
+      key: isNewRecord,
       value: true,
-      enumerable: false,
-      writable: true,
-      configurable: true,
+    }];
```
cyjake, line 117:
```js
Object.assign(this, {
  [raw]: {},
  [rawSaved]: {},
  [rawUnset]: new Set(),
  [rawPrevious]: {},
  isNewRecord: true,
});
```

lib/bone.js
```diff
@@ -144,17 +163,57 @@ class Bone {
...
+    this[rawPrevious][key] = value;
+  }
+
+  isNewRecord() {
+    return this[isNewRecord];
+  }
+
+  setIsNewRecord(value) {
+    this[isNewRecord] = value;
+  }
```
cyjake, line 214: just a second thought, is turning `isNewRecord` into symbol really necessary?

JimmyDaddy: if not so it will be serialized while calling `toJSON` or `toObject`

PR #87, opened 2021-02-19
feat: hasAttribute

JimmyDaddy:
```js

bone.hasAttribute(key)

Bone.hasAttribute(key)
```


test/integration/suite/basics.test.js
```diff
@@ -162,6 +183,8 @@ describe('=> Attributes', function() {
     post.title = 'MHW';
     await post.save();
     assert.deepEqual(post.previousChanges('title'), { title: [ 'Untitled', 'MHW' ] });
+    // should return {} if key not existed
```
cyjake, line 186: 'if key not exists' or 'if key does not exist'

PR #90, opened 2021-02-23
Fix/sequelize mode destroy hooks

JimmyDaddy:
see https://github.com/sequelize/sequelize/blob/4063c2ab627ad57919d5b45cc7755f077a69fa5e/lib/model.js#L2895  before(after)BulkDestroy

destroy with `individualHooks=true` need to put instance in context during hooks executing

### DIFF
--- a/lib/adapters/sequelize.js
+++ b/lib/adapters/sequelize.js
@@ -139,22 +139,22 @@ module.exports = Bone => {
     //   throw new Error('unimplemented');
     // }
 
-    static async destroy(options = {}, runOpt = {}) {
+    static async destroy(options = {}) {
       const { where, individualHooks } = options;
       if (individualHooks) {
         const instances = await this.find(where);
         if (instances.length) {
-          return await Promise.all(instances.map(async (instance) => await instance.destroy(options, runOpt)));
+          return await Promise.all(instances.map((instance) => instance.destroy(options)));
         }
       } else {
-        return await this.bulkDestroy(options, runOpt);
+        return await this.bulkDestroy(options);
       }
     }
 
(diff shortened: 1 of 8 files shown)


lib/adapters/sequelize.js
```diff
@@ -139,7 +139,20 @@ module.exports = Bone => {
     //   throw new Error('unimplemented');
     // }
 
-    static async destroy(options = {}) {
+    static async destroy(options = {}, runOpt = {}) {
+      const { where, individualHooks } = options;
+      if (individualHooks) {
+        const instances = await this.find(where);
+        if (instances.length) {
+          return await Promise.all(instances.map(async (instance) => await instance.destroy(options, runOpt)));
```
cyjake, line 147:
```suggestion
          return await Promise.all(instances.map((instance) => instance.destroy(options, runOpt)));
```

async/await can be omitted here since there's only one promise to wait

lib/adapters/sequelize.js
```diff
@@ -139,7 +139,20 @@ module.exports = Bone => {
     //   throw new Error('unimplemented');
     // }
 
-    static async destroy(options = {}) {
+    static async destroy(options = {}, runOpt = {}) {
```
cyjake, line 142: It seems sequelize does not separate options like options.hooks into another parameter. Shouldn't we stick with their api design?

cyjake: https://sequelize.org/master/class/lib/model.js~Model.html#static-method-destroy

lib/adapters/sequelize.js
```diff
@@ -374,7 +387,7 @@ module.exports = Bone => {
       });
     }
 
-    async destroy(options = {}, runOpt) {
+    async destroy(options = {}, runOpt = {}) {
```
cyjake, line 390: This one as well. I'd suggest keeping `options` only

cyjake (approved):
+1

PR #92, opened 2021-03-01
feat: support multiple order rules in one single string or one-dimensional array

### LINKED ISSUE #86: [FEAT]: support multiple order rules in single string
eg.
```js

model.findAll({
   order: 'created_at asc, id desc',
})
```

JimmyDaddy:
* support orders: fixed #86
```js
 Model.order([ 'created desc', 'id asc' ]);
 Model.order([ [ 'title','desc' ], [ 'id', 'asc' ] ]); 
 Model.order([ 'title', 'desc' ]);
```

* fix where conditions contain empty array cause incorrect SQL 
```js

User.where({
  id: [ ],
});
/**
 * error: SELECT * FROM `users` WHERE `deleted_at` is NULL AND `id` in ()
 * correct: SELECT * FROM `users` WHERE `deleted_at` is NULL 
 * */

```

### DIFF
--- a/lib/drivers/abstract/spellbook.js
+++ b/lib/drivers/abstract/spellbook.js
@@ -26,10 +26,11 @@ function findModel(spell, qualifiers) {
  * @param {Object[]} orders
  */
 function formatOrders(spell, orders) {
-  return orders.map(([token, order]) => {
+  const rs = orders.map(([token, order]) => {
     const column = formatColumn(spell, token);
     return order == 'desc' ? `${column} DESC` : column;
   });
+  return rs;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


lib/drivers/abstract/spellbook.js
```diff
@@ -109,8 +109,11 @@ function formatLiteral(spell, ast) {
 
   if (value == null) {
     return 'NULL';
+  } else if (Array.isArray(value)) {
+    if (value.length) return `(${value.map(() => '?').join(', ')})`;
+    return '';
```
cyjake, line 114: What bug was this change intend to fix?

JimmyDaddy: Empty array in where conditions may cause SQL error
```js

User.where({
  id: [ ],
});
/**
 * error: SELECT * FROM `users` WHERE `deleted_at` is NULL AND `id` in ()
 * correct: SELECT * FROM `users` WHERE `deleted_at` is NULL 
 * */
```

cyjake (approved):
+1

Issue #96, opened 2021-03-05
[feat] sequelize mode scope API


JimmyDaddy:
see https://sequelize.org/master/manual/scopes.html

PR #97, opened 2021-03-05
Feat/scope

JimmyDaddy:
see #96 

### scope [sequelize mode only]?
* model.init with scopes and default scope
```js

Model.init(attributes, {
  defaultScope: { where: { id: 1 } },
  scopes: {
     highLevel: {
        where: {
           level: {
               $gt: 10
           }
        }
     },
     gmail: {
        where: {
          email: {
            $like: '%gmail%'
          }
        }
     }, 
 }
})

User.where({ nickname: 'OldHunter' })
// SELECT * FROM "users" WHERE "nickname" = 'OldHunter' AND "id" = 1 AND "gmt_deleted" IS NULL 
User.scope("highLevel").where({  nickname: 'OldHunter' })
// SELECT * FROM "users" WHERE "nickname" = 'OldHunter' AND "level" > 10 AND "gmt_deleted" IS NULL

```
* pass scopes through array
```js

// array should work
  const scopes = [{
    where: {
      id: 1,
    },
  }, {
    where: {
      author_id: 1
    }
  }];

MyPost.scope(scopes).where({ title: 'New Post' })
//'SELECT * FROM "articles" WHERE "title" = \'New Post\' AND "id" = 1 AND "author_id" = 1 AND "gmt_deleted" IS NULL'
```
* function scope
```js

Model.init(attributes, {
    custom: (order, limit, level) => {
        return {
          where: {
            level,
          },
          order,
          limit,
        };
      }
  }
})


User.scope('custom', 'id asc', 2, 10).where({ nickname: 'OldHunter' })
// SELECT * FROM "users" WHERE "nickname" = 'OldHunter' AND "level" = 10 AND "gmt_deleted" IS NULL ORDER BY "id" LIMIT 2

User.scope(() => {
   return {
     where: {
         level: random()
     }
   }
});

// SELECT * FROM "users" WHERE  "level" = "a random number" AND "gmt_deleted" IS NULL
```

* add scope
```js

User.addScope('name', scope); // scope can be object or function

```

### unparanoid
```js
User.where({ nickname: 'OldHunter' })
// SELECT * FROM "users" WHERE "nickname" = 'OldHunter' AND "gmt_deleted" IS NULL 

User.where({ nickname: 'OldHunter' }).unparanoid
// SELECT * FROM "users" WHERE "nickname" = 'OldHunter'

```

### DIFF
--- a/lib/adapters/sequelize.js
+++ b/lib/adapters/sequelize.js
@@ -338,7 +338,7 @@ module.exports = Bone => {
       }
     }
 
-    // proxy to class.update({ individualHooks=false }) see https://github.com/sequelize/sequelize/blob/4063c2ab627ad57919d5b45cc7755f077a69fa5e/lib/model.js#L2895  before(after)BulkUpdate
+    // proxy to class.update({ individualHooks=false }) see https://github.com/sequelize/sequelize/blob/4063c2ab627ad57919d5b45cc7755f077a69fa5e/lib/model.js#L3083  before(after)BulkUpdate
     static bulkUpdate(values, options = {}) {
       const { where, paranoid = false, validate } = options;
       const whereConditions = where || {};
--- a/lib/adapters/sequelize.js
+++ b/lib/adapters/sequelize.js
@@ -322,10 +322,26 @@ module.exports = Bone => {
       return spell;
     }
 
-    static update(values, options = {}) {
+    static async update(values, options = {}) {
+      const { where, paranoid = false, validate, individualHooks } = options;
+      if (individualHooks) {
+        let findSpell = this.find(where);
+        if (!paranoid) {
+          findSpell = findSpell.unscoped;
+        }
+        const instances = await findSpell;
+        if (instances.length) {
+          return await Promise.all(instances.map((instance) => instance.update(options, { validate })));
+        }
+      } else {
+        return this.bulkUpdate(values, options);
+      }
+    }
+
+    static bulkUpdate(values, options = {}) {
       const { where, paranoid = false, validate } = options;
       const whereConditions = where || {};
-      const spell = super.update(whereConditions, values, { validate });
+      const spell = super.update(whereConditions, values, { validate, hooks: false });
       if (!paranoid) return spell.unscoped;
       return spell;
     }
--- a/lib/adapters/sequelize.js
+++ b/lib/adapters/sequelize.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const { setupSingleHook } = require('../setup_hooks');
-const { compose } = require('../utils');
+const { compose, isPlainObject } = require('../utils');
 
 function translateOptions(spell, options) {
   const { attributes, where, group, order, offset, limit, include } = options;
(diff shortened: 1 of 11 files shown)


cyjake:
Didn't notice much of `scope` use in skylark until today. In my humble opinion, I'd say the part of sequelize api design is way too clumsy. Personally I'd prefer static methods that extends the `spell` created by `static._find()`, such as:

```js
class User extends Bone {
  static defaultScope(spell) {
    return spell.$where({ deletedAt: null });
  }

  static highLevel(spell) {
    return spell.$where('level >= 10');
}
```

About the `Model.scope(...options)` part, not very fond of this kind of api either. Was it required in our existing code?

cyjake:
Anyway, leaving this at the sequelize adapter is fine by me. Nice work.

lib/adapters/sequelize.js
```diff
@@ -41,24 +41,124 @@ function translateOptions(spell, options) {
   if (offset) spell.$offset(offset);
 }
 
+const setScopeToSpell = (scope) => (spell) => {
```
cyjake, line 44: I smell redux middleware

lib/adapters/sequelize.js
```diff
@@ -41,24 +41,124 @@ function translateOptions(spell, options) {
...
+    }
+  }
+  return merged;
+};
+
+/**
+ * parse scope
+ * @param {Object} scope
+ * @param {Spell} spell
+ * @returns
```
cyjake, line 76: - return type not defined
- a few examples about `parseScope()` arguments would be better

lib/adapters/sequelize.js
```diff
@@ -41,24 +41,124 @@ function translateOptions(spell, options) {
...
 // https://sequelize.org/master/class/lib/model.js~Model.html
 // https://sequelize.org/master/manual/model-querying-finders.html
 module.exports = Bone => {
   return class Spine extends Bone {
 
+    // store all configured scopes
+    // protect
+    static _scopes = {}
+    // scope
+    static _scope = null;
```
cyjake, line 104: does `_scope` serve as the default scope?

lib/adapters/sequelize.js
```diff
@@ -41,24 +41,124 @@ function translateOptions(spell, options) {
...
+}
+
 // https://sequelize.org/master/class/lib/model.js~Model.html
 // https://sequelize.org/master/manual/model-querying-finders.html
 module.exports = Bone => {
   return class Spine extends Bone {
 
+    // store all configured scopes
+    // protect
+    static _scopes = {}
```
cyjake, line 102: `@protected`

are all scopes stored in `_scopes` named?

lib/adapters/sequelize.js
```diff
@@ -85,7 +195,7 @@ module.exports = Bone => {
 
       const { where } = options;
       let spell = this._find(where)[`$${func}`](name);
-      if (options.paranoid === false) return spell.unscoped;
+      if (options.paranoid === false) return spell.unparanoid;
```
cyjake, line 198: I still am a bit hesitant about this `unparanoid` thing. Perhaps we can make all `_find()` calls unscoped and add `where({ deleted_at: null })` accordingly if paranoid is `false`?

The reason about my hesitation is, `unparanoid` is not even a word. `unscoped` is not that legit either but it still has appearances on the web. Perhaps `inparanoid`? 

Anyway, maybe we should have this discussed on Monday at  work.

lib/adapters/sequelize.js
```diff
@@ -301,34 +411,26 @@ module.exports = Bone => {
...
-    }
-
-    static unscoped() {
-      const spell = this._find();
-      spell.scopes = [];
-      return spell;
+    // TODO or TRUNCATE TABLE tableName ?
+    static truncate(options) {
+      const opt = options || {};
+      return this.remove({}, true, opt);
```
cyjake, line 426: I think a specific `TRUNCATE` sql is necessary. The actual query can be implemented at the driver schema.

lib/adapters/sequelize.js
JimmyDaddy: `_scopes` store all named scopes, it can be inherited to subclass. `_scope` is kind of ”class static private variable“,  we can modify it through `setScope`, and it will be used by `Spell`

lib/adapters/sequelize.js
JimmyDaddy: 😂 Because [Spell accept function scope](https://github.com/cyjake/leoric/blob/1fff0b95243db4ba11f766f67c7dffece742a201/lib/drivers/abstract/spellbook.js#L692)

PR #102, opened 2021-03-18
fix: static create validated twice

JimmyDaddy:


lib/bone.js
```diff
@@ -1077,9 +1077,6 @@ class Bone {
    * @return {Spell}
    */
   static create(values, opts = {}) {
-    if (opts.validate !== false) {
-      this._validateAttributes(values);
-    }
```
cyjake, line 1082: just a second thought, how about still validate here, but turn validation off in `instance.create(values, { validate: false })` L1085

JimmyDaddy: that will make validator can't access `this`:
```js

Model.init({
  name: {
     validate: {
        custom: function () {
          return  this.name.length > 10; // this.name will throw error
        }
     }
  }
})
```

PR #104, opened 2021-03-23
fix: update, sequelize mode get API, destroy compitable

JimmyDaddy:
fixed:

* `update` values should be immutable
* sequelize get
* sequelize `destroy` returns value
* sequelize `findByPk` trans options
* sequelize:  get unset attribute should not throw error when calling `getDataValues(key)`
* refactor serialize, multiple inheritance is no longer considered
* realm.escape
* `remove` should update instance in paranoid mode


lib/adapters/sequelize.js
```diff
@@ -537,8 +538,16 @@ module.exports = Bone => {
...
+    /**
+     *
+     *
+     * @param {*} [options={}]
+     * @returns
+     */
     async destroy(options = {}) {
-      return await this.remove(options.force, { ...options, hooks: false });
+      const removeResult = await this._remove(options.force, { ...options });
+      if (options.force) return removeResult;
```
JimmyDaddy, line 549: force delete return affects row, otherwise return instance

lib/bone.js
```diff
@@ -1195,17 +1194,20 @@ class Bone {
    */
   static update(conditions, values = {}, queryOption = {}) {
     const { attributes } = this;
+    // values should be immutable
+    const data = Object.assign({}, values);
```
JimmyDaddy, line 1198: values should be imuutable

lib/validator.js
```diff
@@ -82,6 +84,7 @@ function executeValidator(ctx, name, attribute, setValue) {
     } catch (err) {
       // revert value (instance only)
       if (ctx.raw && needToRevert) ctx.raw[field] = originValue;
+      err.name = ERR_NAME;
```
JimmyDaddy, line 87: `error.name` should be identifiable, thus developer can use it for properly response.

test/integration/suite/associations.test.js
```diff
@@ -156,9 +156,9 @@ describe('scattered associations', function() {
 
   it('should return duplicated records', async function() {
     const posts = await Post.all.with('comments');
-    assert.deepEqual(Array.from(posts[0].comments, comment => comment.content), [
+    assert.deepEqual(Array.from(posts[0].comments, comment => comment.content).sort(), [
       'Abandon your foolish request!',
       "Now you'll join them"
-    ]);
+    ].sort());
```
cyjake, line 162: this one is unnecessary, `Array.from(posts[0].comments, comment => comment.content).sort()` sourts elements in alphabetical order already

lib/validator.js
```diff
@@ -104,7 +107,14 @@ function executeValidator(ctx, name, attribute, setValue) {
     }
     let callableArgs = [ String(value) ];
     callableArgs = callableArgs.concat(args);
-    if (!validator.apply(ctx, callableArgs)) throw new LeoricValidateError(name, field, msg);
+    let result;
+    try {
+      result = validator.apply(ctx, callableArgs);
+    } catch (e) {
+      e.name = ERR_NAME;
+      throw e;
```
cyjake, line 115: An unspoken convention I've been using in leoric is that, `event` is simplified as `e`, and `error` is simplified as `err`, can we follow this convention here?

btw, does this change mean when `validator.apply()` fails, it might return false, but also might throw an error? Sounds a bit weird.

JimmyDaddy: build-in validator only returns boolean, my fault

lib/adapters/sequelize.js
```diff
@@ -650,6 +651,24 @@ module.exports = Bone => {
...
     }
 
+    /**
+     * dataValues, request ctx or other config those developer mounted on Model should not be serialized
+     * @param {Array<string>} [exceptKeys=[]] [ 'ctx', 'config' ]
+     * @returns {Object}
+     */
+    toJSON(exceptKeys = []) {
+      return super.toJSON([ 'dataValues' ].concat(exceptKeys));
+    }
```
cyjake, line 661: this one might interfere with the standard behavior, reference https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior

cyjake: consider setting dataValues to `enumerable: false`

lib/bone.js
```diff
@@ -429,16 +429,16 @@ class Bone {
...
    * @return {Object}
    */
-  toJSON() {
+  toJSON(exceptKeys = []) {
     const obj = {};
 
     const names = getPropertyNames(this);
     for (const key of names) {
-      if (this.#rawUnset.has(key) || key === 'isNewRecord') continue;
+      if (this.#rawUnset.has(key) || key === 'isNewRecord' || exceptKeys.includes(key)) continue;
```
cyjake, line 437: this one too. I think we shall set isNewRecord `enumerable: false` as well

lib/adapters/sequelize.js
JimmyDaddy: `enumerable  = false` cant filter, `isNewRecord` is non-enumerable already.

test/integration/suite/basics.test.js
```diff
@@ -446,20 +463,20 @@ describe('=> Integration', function() {
     assert.deepEqual(Object.keys(comments[0]), Object.keys(Comment.attributes));
   });
```
JimmyDaddy, line 465: multiple inheritance is no longer considered in serialize

test/integration/suite/basics.test.js
```diff
@@ -446,20 +463,20 @@ describe('=> Integration', function() {
...
-      }
-    }
-    const post = await CustomPost.findOne({ title: 'New Post' });
-    const json = post.toJSON();
-    assert.equal(json.customProperty, 'customProperty');
-    const obj = post.toObject();
-    assert.equal(obj.customProperty, 'customProperty');
-
-  });
+  // it('bone.toJSON() and bone.toObject() should work when multiple extends', async () => {
```
JimmyDaddy, line 466: multiple inheritance is no longer considered in serialize

PR #105, opened 2021-04-15
Fix/transaction fix

JimmyDaddy:
fix:  Transaction integration with Bone API

### DIFF
--- a/lib/bone.js
+++ b/lib/bone.js
@@ -1065,16 +1065,14 @@ class Bone {
    * @return {Spell}
    */
   static _find(conditions, ...values) {
-    let spell;
     const conditionsType = typeof conditions;
     // find(1)
     // find([ 1, 2, 3 ])
+    let spell = new Spell(this);
     if (Array.isArray(conditions) || conditionsType == 'number') {
       if (values.length == 1 && typeof values[0] === 'object') {
         // pass options to spell
         spell = new Spell(this, values[0]);
-      } else {
-        spell = new Spell(this);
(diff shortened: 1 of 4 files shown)


lib/drivers/abstract/schema.js
```diff
@@ -168,7 +168,10 @@ module.exports = {
       VALUES ${placeholders.join(', ')}
     `);
     if (returning) sql = `${sql} RETURNING ${returning.join(', ')}`;
-
-    return await this.query(sql, values);
+    let queryOptions = {};
+    if (opts.connection) {
+      queryOptions.connection = opts.connection;
+    }
+    return await this.query(sql, values, queryOptions);
```
cyjake, line 175:
```suggestion
    const { connection } = opts;
    return await this.query(sql, values, { connection });
```

lib/bone.js
```diff
@@ -1070,7 +1070,12 @@ class Bone {
     // find(1)
     // find([ 1, 2, 3 ])
     if (Array.isArray(conditions) || conditionsType == 'number') {
-      spell = new Spell(this);
+      if (values.length == 1 && typeof values[0] === 'object') {
+        // pass options to spell
+        spell = new Spell(this, values[0]);
+      } else {
+        spell = new Spell(this);
+      }
```
cyjake, line 1078: similar checks already occur in the else branch at L1081, I think the outer if needs some refactor

cyjake (approved):

PR #106, opened 2021-04-16
release: v1.4.0-alpha.1

cyjake:
* feat: support multiple order rules in one single string or one-dimensional array (#92)
  * feat: `realm.raw('SELECT ...')` and `Model.raw('SELECT ...')`
  * feat: `Model.truncate()` now uses TRUNCATE if possible
  * feat: `Model.find().optimizerHints('SET_VAR(foreign_key_checks=OFF)')`
  * fix: make sure connection is passed around in all queries carried out within transaction
  * docs: migrations, validations, hooks, and sequelize adapter

### DIFF
--- a/lib/bone.js
+++ b/lib/bone.js
@@ -608,32 +608,32 @@ class Bone {
   /**
    * @public
    * Persist changes on current instance back to database with `UPDATE`.
-   * @param {*} opts
+   * @param {*} changes
    * @returns
    * @memberof Bone
    */
-  update(opts, queryOpts = {}) {
-    return this._update(opts, queryOpts);
+  update(changes, options = {}) {
+    return this._update(changes, options);
   }
   /**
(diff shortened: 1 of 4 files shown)


lib/bone.js
```diff
@@ -1066,27 +1066,26 @@ class Bone {
...
-      }
+      // find(1)
+      // find([ 1, 2, 3 ])
       spell.$where({ [this.primaryKey]: conditions });
+    } else if (typeof conditions === 'object' && options) {
+      // find({}, { offset: 1, limit: 1, ...etc })
+      spell.$where(conditions);
+    } else if (conditions) {
+      // find('title = ?', 'foo')
+      spell.$where(conditions, ...values);
```
cyjake, line 1081: @JimmyDaddy refactored a bit

PR #108, opened 2021-04-20
feat: batchUpsert fix#107

JimmyDaddy:
updateOnDuplicate
   * see https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html
   * see https://www.postgresql.org/docs/9.5/sql-insert.html
   * see https://www.sqlite.org/lang_UPSERT.html
   
```SQL
## Postgres SQLite
INSERT INTO `tableName` (`name`)  VALUES ('foo'),('bar') ON CONFLICT (`name`)  DO UPDATE SET `name`=EXCLUDED.`name`;

## MySQL
INSERT INTO `tableName` (`name`) VALUES ('foo'),('bar') ON DUPLICATE KEY UPDATE `name`=VALUES(`name`);
```

```js
// MySQL
User.bulkCreate(records, { updateOnDuplicate: true });
User.bulkCreate(records, { updateOnDuplicate: ['name', 'email'] });

// Postgres SQLite
User.bulkCreate(records, { updateOnDuplicate: true });
User.bulkCreate(records, { updateOnDuplicate: ['name', 'email'] });
User.bulkCreate(records, { updateOnDuplicate: ['name', 'email'], uniqueKeys: ['email'] });
```


lib/drivers/abstract/schema.js
```diff
@@ -163,10 +169,77 @@ module.exports = {
...
+            for (const field of opts.conflictKeys) {
+              const attribute = escapeId(unalias(field));
+              conflictKeys.push(attribute);
+            }
+          } else {
+            conflictKeys.push(escapeId(unalias(opts.conflictKeys)));
+          }
+        } else {
+          // conflict_target must be unique
+          // get all unique keys
```
JimmyDaddy, line 205: How to get all unique indexes?

test/models/user.js
```diff
@@ -25,7 +25,10 @@ class User extends Bone {
 
 // test init
 User.init({
-  id: DataTypes.BIGINT,
+  id: {
+    primaryKey: true,
```
JimmyDaddy, line 29: Cant detect primary key automatically

lib/drivers/abstract/schema.js
```diff
@@ -128,13 +128,74 @@ module.exports = {
...
+          if (att.unique) {
+            uniqueKeys.push(escapeId(att.columnName));
+            break;
+          }
+        }
+      }
+      const { primaryKey } = opts;
+      if (!uniqueKeys.length) uniqueKeys.push(escapeId(primaryKey));
+      // default use id as primary key
+      if (!uniqueKeys.length) uniqueKeys.push(escapeId('id'));
```
cyjake, line 170: Looks like it's possible to push `id` into `uniqueKeys` twice? Will that cause sql error?

lib/drivers/abstract/schema.js
```diff
@@ -128,13 +128,74 @@ module.exports = {
...
+    }
+
+    if (Array.isArray(opts.updateOnDuplicate) && opts.updateOnDuplicate.length) {
+      for (const field of opts.updateOnDuplicate) {
+        const attribute = escapeId(field);
+        updateKeys.push(`${attribute}=EXCLUDED.${attribute}`);
+      }
+    } else {
+      for (const attribute of attributes) {
+        updateKeys.push(`${attribute.columnName}=EXCLUDED.${attribute.columnName}`);
```
cyjake, line 180: `EXCLUDED` should be commented here, since it's PostgreSQL only.

cyjake: Theoretically speaking, the default vendor in `lib/drivers/abstract/schema.js` should be MySQL or ANSI-SQL

lib/drivers/abstract/schema.js
```diff
@@ -128,13 +128,74 @@ module.exports = {
...
+
+    if (opts.uniqueKeys) {
+      if (Array.isArray(opts.uniqueKeys) && opts.uniqueKeys.length) {
+        for (const field of opts.uniqueKeys) {
+          const attribute = escapeId(field);
+          uniqueKeys.push(attribute);
+        }
+      } else {
+        uniqueKeys.push(escapeId(opts.uniqueKeys));
+      }
```
cyjake, line 153:
```suggestion
     for (const field of [].concat(opts.uniqueKeys)) {
       uniqueKeys.push(escapeId(field));
      }
```

test/unit/drivers/sqlite.test.js
```diff
@@ -73,7 +73,7 @@ describe('=> SQLite driver', () => {
       id: { type: BIGINT, primaryKey: true, autoIncrement: true },
       title: { type: STRING, allowNull: false },
     });
-    await driver.query(`INSERT INTO notes (id, title) VALUES (42, 'Untitled')`);
+    await driver.query('INSERT INTO notes (id, title) VALUES (42, \'Untitled\')');
```
cyjake, line 76: These changes are a bit unnecessary. Can the corresponding eslint rule be adjusted?

lib/drivers/mysql/schema.js
```diff
@@ -77,4 +77,24 @@ module.exports = {
...
+    let attrStr;
+    const { escapeId } = this;
+    const { attributes } = opts;
+    if (Array.isArray(opts.updateOnDuplicate) && opts.updateOnDuplicate.length) {
+      attrStr = opts.updateOnDuplicate.map((field) => {
+        const attribute = escapeId(field);
+        return `${attribute}=VALUES(${attribute})`;
+      }).join(',');
+    } else {
+      attrStr = attributes.map((attribute) => `${attribute.columnName}=VALUES(${attribute.columnName})`).join(',');
```
cyjake, line 96: `columnName` should be escaped as well

cyjake, line 96:
```suggestion
      attrStr = attributes.map((attribute) => {
        const columnStr = escapeId(attribute.columnName);
        return `${columnStr}=VALUES(${columnStr})`
      }).join(',');
```

lib/drivers/abstract/schema.js
cyjake: seems SQLite shares same syntax as PostgreSQL, agreed on current arrangement.

PR #111, opened 2021-05-18
revise instructions for installing Jekyll

DanielKehoe:
Previous instructions were outdated as Homebrew uses a different directory on Apple Silicon. Easier to link to current instructions on the Jekyll website, plus give a link to a complete guide at mac.install.guide that explains the pitfalls of using Homebrew to install Ruby, and point to Moncef Belyamani's Ruby setup scripts. Hope this helps!


cyjake:
Thanks for the update!

PR #118, opened 2021-06-23
refactor: spell_insert

JimmyDaddy:
* refactor spellBook insert commands, `upsert` and `bulkInsert` base on `insert`
* fix `find` API in Sequelize mode
* `Bone.upsert`


JimmyDaddy:
ci: https://github.com/JimmyDaddy/leoric/runs/2897478787

lib/bone.js
```diff
@@ -1339,6 +1388,7 @@ class Bone {
       if (attributes.hasOwnProperty(snakeCase(name))) {
         timestamps[name] = snakeCase(name);
       }
+      if (attributes.hasOwnProperty(LEGACY_TIMESTAMP_MAP[name])) timestamps[name] = LEGACY_TIMESTAMP_MAP[name];
```
JimmyDaddy, line 1391: `Bone.init` isn't an async function , cannot  call `querySchemaInfo` in it, so take a `LEGACY_TIMESTAMP_MAP ` patch here too.

lib/drivers/abstract/spellbook.js
```diff
@@ -756,6 +840,7 @@ module.exports = {
 
     switch (spell.command) {
       case 'insert':
+      case 'bulkInsert':
```
JimmyDaddy, line 843: Whether merges `bulkInsert` and `insert` actions  are unified as `insert`?

cyjake (approved):
+1

PR #120, opened 2021-06-24
refactor: do not subclass by default

cyjake:
- gather around glue code for legacy timestamp

### DIFF
--- a/test/unit/realm.test.js
+++ b/test/unit/realm.test.js
@@ -37,12 +37,12 @@ describe('=> Realm', () => {
     assert.ok(DataTypes);
   });
 
-  it('should subclass Bone unless specified by user', async () => {
-    const realm = new Realm();
-    assert.ok(realm.Bone.prototype instanceof Bone);
-
-    const arda = new Realm({ Bone });
+  it('should not subclass Bone unless asked specifically', async () => {
+    const arda = new Realm();
     assert.equal(arda.Bone, Bone);
+
+    const realm = new Realm({ subclass: true });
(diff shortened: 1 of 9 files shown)


test/unit/realm.test.js
```diff
@@ -37,12 +37,12 @@ describe('=> Realm', () => {
...
-    const realm = new Realm();
-    assert.ok(realm.Bone.prototype instanceof Bone);
-
-    const arda = new Realm({ Bone });
+  it('should not subclass Bone unless asked specifically', async () => {
+    const arda = new Realm();
     assert.equal(arda.Bone, Bone);
+
+    const realm = new Realm({ subclass: true });
+    assert.ok(realm.Bone.prototype instanceof Bone);
```
cyjake, line 45: egg-orm will turn `opts.subclass` on by default, which (theoretically) makes connecting to multiple databases simultaneously possible.

JimmyDaddy: `subclass` is not easy to understand, how about change another name,  for example: `singleConnection`?

cyjake: There always is only one connection pool for realm instance. When [multiple `datasources`](https://github.com/eggjs/egg-orm/blob/master/lib/loader.js#L33) is specified in egg-orm, egg-orm tries to instantiate multiple realms to retain multiple connection pools with the config specified in `datasources` array.

We made this possible by subclassing Bone. But this feature is seldom used (at least for now) hence I intend to not make this the default behavior.

PR #125, opened 2021-06-28
Feat/complete bone/spine.restore and Bone API type definitions

JimmyDaddy:


types/index.d.ts
```diff
@@ -476,6 +498,46 @@ declare class Bone {
...
+   * reload instance
+   */
+  reload(): Promise<Bone>;
+
+  /**
+   * delete row
+   * @param forceDelete force update
+   * @param opts query options
+   */
+  remove(forceDelete?: boolean, opts?: QueryOptions): Spell & Promise<number>;
```
cyjake, line 534: overloaded methods with more arguments should be put before the generic one. https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#function-overloads

types/index.d.ts
```diff
@@ -383,12 +404,12 @@ declare class Bone {
...
    * UPDATE rows.
    */
-  static update(whereConditions: WhereConditions): Spell & Promise<number>;
+  static update(whereConditions: WhereConditions, values?: Object, opts?: QueryOptions): Spell & Promise<number>;
 
   /**
    * Remove rows. If soft delete is applied, an UPDATE query is performed instead of DELETing records directly. Set `forceDelete` to true to force a `DELETE` query.
    */
-  static remove(whereConditions: WhereConditions, forceDelete?: boolean): Spell & Promise<number>;
+  static remove(whereConditions: WhereConditions, forceDelete?: boolean, opt?: QueryOptions): Spell & Promise<number>;
```
cyjake, line 412:
```suggestion
  static remove(whereConditions: WhereConditions, forceDelete?: boolean, opts?: QueryOptions): Spell & Promise<number>;
```

PR #127, opened 2021-07-01
fix: leave the getter properties defined in `class` syntax as is

cyjake:
```js
class Foo {
  get a() {}
}
```

`foo.a` is not enumerable by default, Leoric should leave it that way. If users want them to be enumerable, they need to redefine them explicitly

### DIFF
--- a/src/bone.js
+++ b/src/bone.js
@@ -13,7 +13,7 @@ const Collection = require('./collection');
 const Spell = require('./spell');
 const { capitalize, camelCase, snakeCase } = require('./utils/string');
 const { setupHooks } = require('./setup_hooks');
-const { getPropertyNames, logger } = require('./utils/index');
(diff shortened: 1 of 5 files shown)


cyjake:
This is yet another major differences between `class` and `Model.define({ getterMethods })`, will be noted in https://www.yuque.com/leoric/blog/op7b06

cyjake:
cc @chenhui5416

src/bone.js
```diff
@@ -487,14 +490,13 @@ class Bone {
   toObject() {
     const obj = {};
 
-    const names = getPropertyNames(this);
-    for (const key of names) {
+    for (const key in this) {
```
cyjake, line 493: `for..in` does basically the same thing as `getPropertyNames()` util

JimmyDaddy (approved):

PR #130, opened 2021-07-03
WIP: tweak find to query multiple values

xudafeng:


src/adapters/sequelize.js
```diff
@@ -314,7 +314,7 @@ module.exports = Bone => {
     }
 
     static find(options = {}) {
-      return this.findOne(options);
+      return this.findAll(options);
```
cyjake, line 317: Can't do that. Sequelize has `Model.find = Model.findOne` in its source code and we, no matter purposefully or not,  have been using this behavior for a while. Changing the return type breaks both sequelize API adapter and our existing code.

PR #131, opened 2021-07-05
feat: support class static attributes

cyjake:
```js
class User extends Bone {
  static attributes = {}
  static beforeCreate(obj) {
  }
  static initialize() {
    this.hasMany('posts');
  }
}
```

- deprecated `Bone.describe()`, migrating to `Bone.initialize()`.
- renamed `Bone.aliasName` to `Bone.tableAlias`, which is an internal property

### DIFF
--- a/index.js
+++ b/index.js
@@ -80,7 +80,7 @@ async function loadModels(Bone, models, opts) {
 
   for (const model of models) {
     model.describe();
-    model.didLoad();
+    model.initialize();
   }
 }
 
--- a/index.js
+++ b/index.js
@@ -80,6 +80,7 @@ async function loadModels(Bone, models, opts) {
 
   for (const model of models) {
     model.describe();
+    model.didLoad();
   }
 }
 
--- a/src/bone.js
+++ b/src/bone.js
@@ -917,7 +917,7 @@ class Bone {
     const { attributes, options } = this;
     const attributeMap = {};
     const table = this.table || snakeCase(pluralize(this.name));
-    const aliasName = camelCase(pluralize(this.name || table));
+    const tableAlias = camelCase(pluralize(this.name || table));
 
     this.normalize(attributes);
     for (const name of Object.keys(attributes)) {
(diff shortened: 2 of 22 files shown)


index.js
```diff
@@ -80,6 +80,7 @@ async function loadModels(Bone, models, opts) {
 
   for (const model of models) {
     model.describe();
+    model.didLoad();
```
cyjake, line 83: @JimmyDaddy any suggestions about this name?

src/bone.js
```diff
@@ -941,6 +980,19 @@ class Bone {
    */
   static describe() {}
 
+  /**
+   * Override this method to setup associations, rename attributes, etc.
+   * @deprecated
```
JimmyDaddy, line 985:
```suggestion
```

index.js
JimmyDaddy: how about `schema` or `schemaLoad` ?

src/bone.js
```diff
@@ -896,13 +916,25 @@ class Bone {
     const { Attribute } = this.driver;
     const { attributes, options } = this;
     const attributeMap = {};
+    const table = this.table || snakeCase(pluralize(this.name));
+    const aliasName = camelCase(pluralize(this.name || table));
```
JimmyDaddy, line 920: tableAlias?

index.js
cyjake: It gets called when model is loaded (synchronized with database). `schemaLoad` might be better suited at `realm.schemaLoaded` I think.

JimmyDaddy: But as a developer who use leoric,  this function is used to load my custom settings, `schemaLoaded`  is not easy to understand.

src/bone.js
cyjake: sounds good. updated

index.js
cyjake: agreed on `static initialize()`

JimmyDaddy (approved):

PR #132, opened 2021-07-06
fix: collection convert should handle tddl results as well

cyjake:
- vanilla mysql: `[ { '': { count: 1 } } ]`
- tddl: `[ { [table]: { count: 1 } } ]`

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -134,15 +134,16 @@ function dispatch(spell, rows, fields) {
  */
 function convert(spell, rows, fields) {
   const results = [];
-  const { joins } = spell;
+  const { groups, joins, columns } = spell;
   const { table, tableAlias } = spell.Model;
 
   // await Post.count()
-  if (rows.length <= 1 && spell.columns.length === 1) {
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -134,15 +134,16 @@ function dispatch(spell, rows, fields) {
...
 
   // await Post.count()
-  if (rows.length <= 1 && spell.columns.length === 1) {
-    const { type, value, args } = spell.columns[0];
+  if (rows.length <= 1 && columns.length === 1 && groups.length === 0) {
+    const { type, value, args } = columns[0];
     if (type === 'alias' && args && args[0].type === 'func') {
       const row = rows[0];
-      return row && row[''] && row[''][value] || 0;
+      const result = row && (row[''] || row[table]);
```
JimmyDaddy, line 145: `row[tableAlias || table]`?

cyjake: tableAlias is only used in join query, which won't be active here

JimmyDaddy (approved):

PR #133, opened 2021-07-08
fix: renamed attribute should remain enumerable

cyjake:
even if there is getter method with same name. Also added `Model.Instance` in sequelize adapter

### DIFF
--- a/src/adapters/sequelize.js
+++ b/src/adapters/sequelize.js
@@ -126,6 +126,10 @@ module.exports = Bone => {
       return true;
     }
 
+    static get Instance() {
+      return this;
+    }
+
     /**
      * add scope see https://sequelize.org/master/class/lib/model.js~Model.html#static-method-addScope
      *
--- a/src/bone.js
+++ b/src/bone.js
@@ -935,9 +935,7 @@ class Bone {
       }
     }
 
-    const descriptors = {};
     for (const name in attributes) this.loadAttribute(name);
-    Object.defineProperties(this.prototype, descriptors);
 
     Object.defineProperties(this, looseReadonly({
       timestamps,
--- a/src/bone.js
+++ b/src/bone.js
@@ -936,23 +936,7 @@ class Bone {
     }
 
     const descriptors = {};
-    for (const name in attributes) {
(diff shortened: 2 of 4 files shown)


src/adapters/sequelize.js
```diff
@@ -126,6 +126,10 @@ module.exports = Bone => {
       return true;
     }
 
+    static get Instance() {
+      return this;
+    }
```
cyjake, line 131: added `Model.Instance` alias

test/unit/bone.test.js
```diff
@@ -5,119 +5,178 @@ const { Bone, DataTypes, connect } = require('../..');
...
-      { columnName: 'ssn', columnType: 'varchar', dataType: 'varchar', primaryKey: true },
-    ]);
-    assert.equal(User.primaryKey, 'ssn');
-    assert.equal(User.primaryColumn, 'ssn');
+      User.load([
+        { columnName: 'foo', columnType: 'varchar', dataType: 'varchar' },
+      ]);
+      User.renameAttribute('foo', 'bar');
+      assert.ok(Object.getOwnPropertyDescriptor(User.prototype, 'bar').enumerable);
+    });
```
cyjake, line 159: It's expected that attributes should always present in `toJSON()` or `toObject()` results. Make renamed attribute enumerable as well.

src/bone.js
```diff
@@ -936,23 +936,7 @@ class Bone {
     }
 
     const descriptors = {};
```
JimmyDaddy, line 938:
```suggestion
```

src/bone.js
```diff
@@ -936,23 +936,7 @@ class Bone {
...
-      }, Object.keys(descriptor || {}).reduce((result, key) => {
-        if (descriptor[key] != null) result[key] = descriptor[key];
-        return result;
-      }, {}), {
-        enumerable: true,
-        configurable: true,
-      });
-    }
+    for (const name in attributes) this.loadAttribute(name);
     Object.defineProperties(this.prototype, descriptors);
```
JimmyDaddy, line 940:
```suggestion
```
unused code

cyjake: nice catch

JimmyDaddy (approved):

PR #145, opened 2021-07-21
fix: subclassing data type in dialects

cyjake:
exports of the abstract data types should not be wrapped with invokable by default, which made subclassing like `class Postgres_DATE extends DataTypes.DATE {}` not working as expected because `DataTypes.DATE` was actually a Proxy instance.

This fix defers the invokable wrapping and adds more test cases regarding data types in dialects.


test/unit/drivers/postgres/attribute.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+  it('should support SERIAL', async function() {
+    const attribute=  new Attribute('id', {
+      type: INTEGER,
+      primaryKey: true,
+    });
+    assert.equal(attribute.toSqlString(), '"id" SERIAL PRIMARY KEY');
+  });
+
+  it('should support WITH or WITHOUT TIME ZONE', async function() {
```
JimmyDaddy, line 24: Need a `WITHOUT TIME ZONE` test case.

test/unit/drivers/postgres/attribute.test.js
```diff
@@ -0,0 +1,37 @@
...
+    const attribute=  new Attribute('createdAt', {
+      type: DATE,
+    });
+    assert.equal(attribute.toSqlString(), '"created_at" TIMESTAMP WITH TIME ZONE');
```
JimmyDaddy, line 28:
```suggestion
    assert.equal(attribute.toSqlString(), '"created_at" TIMESTAMP WITH TIME ZONE');
    const attribute1 =  new Attribute('createdAt', {
      type: new DATE(null, false),
    });
    assert.equal(attribute1.toSqlString(), '"created_at" TIMESTAMP WITHOUT TIME ZONE');
```

test/unit/drivers/postgres/attribute.test.js
cyjake: added

test/unit/drivers/postgres/attribute.test.js
cyjake: added a separated test

PR #146, opened 2021-07-24
fix: JSONB in PostgreSQL & JSON as TEXT

### LINKED ISSUE #147: JSON and JSONB
mysql >= 5.7 supports JSON as column data type, which stores JSON data in binary format and will serialize/deserialize automatically when retrieving from or writing into database. 

postgres supports JSON and JSONB, both have JSON data storage supported but the former one only stores them in text format, and the latter stores in binary, which makes more sense and behaves just like the JSON in mysql

currently, in leoric we provide two different types a bit like postgres:

```js
const { JSON, JSONB } = require('leoric').DataTypes;
```

To maximize database compatibility, the JSON type is handled a bit differently by actually storing them in TEXT because older mysql versions and (most importantly) those enterprise databases that is merely compatible with mysql at the protocol level. model attributes with this type will have the text data parsed or stringified at the model level. Hence in the js land the property is an object, and in the database it's still plain text.

the JSONB type however, is just like the JSON in mysql and the JSONB in postgres, the serialize/deserialize work is done by database, and querying with property path (like `Post.where('params.foo.bar = ?', true)`) is in our plan.

In the long run, we'd prefer to deprecating JSON type one day, and switches both to JSON binary.

cyjake:
fixes #147

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -43,14 +43,14 @@ class Collection extends Array {
    * Save the collection. Currently the changes are made concurrently but NOT in a transaction.
    * @returns {Bone[]}
    */
-  save() {
-    if (this.length === 0) return Promise.resolve(this);
+  async save() {
+    if (this.length === 0) return this;
 
     if (this.some(element => !element.save)) {
       throw new Error('Collection contains element that cannot be saved.');
     }
 
-    return Promise.all(this.map(element => element.save()));
+    return await Promise.all(this.map(element => element.save()));
   }
 }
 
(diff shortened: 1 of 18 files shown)


src/collection.js
```diff
@@ -104,26 +104,30 @@ function dispatch(spell, rows, fields) {
...
+      if (!current[qualifier]) current[qualifier] = new Collection();
+      if (!id || current[qualifier].some(item => item[Model.primaryKey] == id)) continue;
+      current[qualifier].push(Model.instantiate(values));
+    } else {
+      current[qualifier] = Object.values(values).some(value => value != null)
+        ? Model.instantiate(values)
+        : null;
+    }
+  }
+}
```
cyjake, line 129: refactored into separate function to reduce indentation levels

src/data_types.js
```diff
@@ -156,33 +156,43 @@ class BOOLEAN extends DataType {
...
+      throw new Error(`invalid blob length: ${length}`);
+    }
     super();
     this.dataType = 'text';
+    this.length = length;
   }
 
   toSqlString() {
-    return this.dataType.toUpperCase();
+    return [ this.length, this.dataType ].join('').toUpperCase();
```
cyjake, line 172: TINYTEXT, TEXT, MEDIUMTEXT, and LONGTEXT

src/data_types.js
```diff
@@ -156,33 +156,43 @@ class BOOLEAN extends DataType {
...
+    }
     super();
-    this.dataType = 'longblob';
+    this.dataType = 'blob';
+    this.length = length;
   }
 
   toSqlString() {
-    return this.dataType.toUpperCase();
+    return [ this.length, this.dataType ].join('').toUpperCase();
```
cyjake, line 187: I'd say mysql is crazy about types. Again, TINYBLOB, BLOB, MEDIUMBLOB, and LONGBLOB

src/data_types.js
```diff
@@ -156,33 +156,43 @@ class BOOLEAN extends DataType {
...
   }
 }
 
 // JSON text type
 class JSON extends DataType {
   constructor() {
     super();
-    this.dataType = 'json';
+    this.dataType = 'text';
```
cyjake, line 195: `dataType` should reflect the true data type of the column in database, hence changed to `text` in this case

src/data_types.js
```diff
@@ -191,14 +201,16 @@ class JSON extends DataType {
...
 class JSONB extends DataType {
   constructor() {
     super();
-    this.dataType = 'jsonb';
+    this.dataType = 'json';
   }
 
   toSqlString() {
-    return 'json';
+    return 'JSON';
```
cyjake, line 213: default to `json`, overridden in the postgres/data_types with `jsonb`

src/drivers/mysql/schema.js
```diff
@@ -77,24 +76,4 @@ module.exports = {
...
-
-    if (Array.isArray(opts.updateOnDuplicate) && opts.updateOnDuplicate.length) {
-      columns = opts.updateOnDuplicate;
-    } else {
-      columns = attributes.map(attribute => attribute.columnName);
-    }
-
-    const sets = columns.map(column => `${escapeId(column)}=VALUES(${escapeId(column)})`);
-    return `ON DUPLICATE KEY UPDATE ${sets.join(', ')}`;
-  },
```
cyjake, line 99: obsolete code removed

test/unit/drivers/sqlite/index.test.js
```diff
@@ -79,6 +79,52 @@ describe('=> SQLite driver', () => {
...
+      cid: 1,
+      name: 'title',
+      type: 'VARCHAR(255)',
+      notnull: 0,
+      dflt_value: null,
+      pk: 0
+    });
+    const result = await driver.query('SELECT * FROM notes');
+    assert.equal(result.rows.length, 1);
+  });
```
cyjake, line 127: test cases added to increase code coverage

JimmyDaddy (approved):

PR #153, opened 2021-08-05
fix: prevent from calling Date.prototype.toJSON

cyjake:
```js
const user = await User.first;
user.toJSON();
user.createdAt  // should be an instance of Date
```


test/integration/suite/basics.test.js
```diff
@@ -400,17 +400,15 @@ describe('=> Integration', function() {
     expect(result).to.contain('New Post');
   });
 
-  it('bone.toJSON()', async function() {
+  it.only('bone.toJSON()', async function() {
```
xudafeng, line 403: only删除

xudafeng (approved):

PR #157, opened 2021-08-11
test: validator unit test

JimmyDaddy:


src/validator.js
```diff
@@ -14,6 +14,15 @@ class LeoricValidateError extends Error {
   }
 }
 
+function regex(str, pattern, modifiers) {
+  str += '';
+  if (Object.prototype.toString.call(pattern).slice(8, -1) !== 'RegExp') {
```
cyjake, line 19: not sure if `instanceof RegExp` is better but it sounds simpler to understand

PR #168, opened 2021-09-05
fix: bulkCreate(...records) should recognize custom setters

cyjake:
att

### DIFF
--- a/src/bone.js
+++ b/src/bone.js
@@ -1313,6 +1313,9 @@ class Bone {
       opts.uniqueKeys = opts.uniqueKeys.map((field) => this.unalias(field));
     }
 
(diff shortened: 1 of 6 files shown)


src/bone.js
```diff
@@ -1313,6 +1313,9 @@ class Bone {
       opts.uniqueKeys = opts.uniqueKeys.map((field) => this.unalias(field));
     }
 
+    // records might change when filter through custom setters
+    records = instances.map(instance => instance.getRaw());
```
JimmyDaddy, line 1317: line 1326 will never be reached, should be removed

cyjake: indeed, removed now.

PR #169, opened 2021-09-05
feat: support STRING.BINARY(length), complete unit test

JimmyDaddy:
attr

### DIFF
--- a/src/data_types.js
+++ b/src/data_types.js
@@ -188,7 +188,7 @@ const LENGTH_VARIANTS = [ 'tiny', '', 'medium', 'long' ];
 class TEXT extends DataType {
   constructor(length = '') {
     if (!LENGTH_VARIANTS.includes(length)) {
-      throw new Error(`invalid blob length: ${length}`);
+      throw new Error(`invalid text length: ${length}`);
     }
     super();
     this.dataType = 'text';
--- a/src/data_types.js
+++ b/src/data_types.js
@@ -17,6 +17,12 @@ class DataType {
     switch (dataType) {
       case 'varchar':
       case 'char':
+      // mysql only
+      case 'binary':
+      // mysql only
+      case 'varbinary':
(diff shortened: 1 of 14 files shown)


src/data_types.js
```diff
@@ -17,6 +17,12 @@ class DataType {
     switch (dataType) {
       case 'varchar':
       case 'char':
+      // mysql only
+      case 'binary':
+      // mysql only
+      case 'varbinary':
+      // postgres only
+      case 'bytea':
```
cyjake, line 25: just wondering, shouldn't we use js `Binary`?

cyjake: The three newly added `dataType`s should be mapped as `BINARY` type. On the other hand, the `jsType` of `BINARY`, which is currently listed in `src/drivers/abstract/attribute.js`, should be added as well, and `Buffer` should be preferred.

cyjake: besides unit tests regarding the change itself, integration tests need to be added as well. Such as defining model with `content: BINARY` or `avatar: BLOB`, or loading model attributes from schema info such as `foo BINARY`.

src/drivers/abstract/attribute.js
```diff
@@ -11,6 +11,9 @@ const { snakeCase } = require('../../utils/string');
 function findJsType(DataTypes, type, dataType) {
   if (type instanceof DataTypes.BOOLEAN) return Boolean;
   if (type instanceof DataTypes.JSON) return JSON;
+  if (type instanceof DataTypes.BINARY || type instanceof DataTypes.VARBINARY || type instanceof DataTypes.BLOB) {
```
cyjake, line 14: `type instanceof DataTypes.VARBINARY` won't be necessary since `type instanceof DataTypes.BINARY` already has this condition covered.

cyjake (approved):

PR #170, opened 2021-09-11
refactor: clean some old code

okbug:


src/expr.js
```diff
@@ -281,29 +277,29 @@ function parseExprList(str, ...values) {
 
   function operator(t) {
     const op = token();
-    if (op.name == 'as') return alias(t);
-    if (op.name == 'between' || op.name == 'not between') {
+    if (op.name === 'as') return alias(t);
+    if (op.name === 'between' || op.name == 'not between') {
```
JimmyDaddy, line 281: op.name === 'not between'. Or ['between','not between'].includes(op.name) ?

src/spell.js
```diff
@@ -463,7 +465,7 @@ function scopeDeletedAt(spell) {
   for (const condition of whereConditions) {
     let found = false;
     walkExpr(condition, ({ type, value }) => {
-      if (type == 'id' && value == deletedAt) {
+      if (type === 'id' && value == deletedAt) {
```
JimmyDaddy, line 468:
```
value === deletedAt
```

okbug: if value is `null` or `undefined`, use `===` will return `false`.

JimmyDaddy (approved):

JimmyDaddy (commented):
Did you run unit test first? It seems there are some test didn't pass, BTW this PR seems skipped github checks

src/spell.js
```diff
@@ -81,9 +81,11 @@ function parseObjectValue(value) {
  * @returns {boolean}
  */
 function isOperatorCondition(condition) {
+  const keys = Object.keys(condition);
```
cyjake, line 84: `Object.keys(condition)` should be called after `isPlainObject(condition)` passes

cyjake: which is the reason that causes tests to fail

PR #171, opened 2021-09-13
refactor: called Object.keys() after isPlainObjects()

okbug:
I havv moved the `isPlainObject` function before `Object.keys()`


cyjake:
Previous changes are merged already, please rebase against origin/master and update your branch

PR #172, opened 2021-09-13
fix: attribute.equals() check should ignore defaultValue

cyjake:
`attribute.defaultValue` currently have two mixed responsibility:
- the default value set in table schema
- the default value to be set when saving records via model methods

### DIFF
--- a/src/drivers/abstract/attribute.js
+++ b/src/drivers/abstract/attribute.js
@@ -106,7 +106,7 @@ class Attribute {
 
   equals(targetAttribute) {
     if (!targetAttribute) return false;
-    const props = [ 'dataType', 'allowNull', 'defaultValue', 'primaryKey' ];
+    const props = [ 'dataType', 'allowNull', 'primaryKey' ];
     for (const prop of props) {
       // SQLite has default value as string even if data type is integer
       if (this[prop] != targetAttribute[prop]) {
--- a/src/drivers/sqlite/pool.js
+++ b/src/drivers/sqlite/pool.js
@@ -1,0 +1,51 @@
+'use strict';
+
+const EventEmitter = require('events');
+const Connection = require('./connection');
+
+class Pool extends EventEmitter {
+  constructor(opts) {
+    super(opts);
+    this.options = {
+      connectionLimit: 10,
+      ...opts,
+      client: opts.client || 'sqlite3',
+    };
+    this.client = require(this.options.client);
+    this.connections = [];
+    this.queue = [];
+  }
+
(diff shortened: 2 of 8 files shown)


src/drivers/abstract/attribute.js
```diff
@@ -106,7 +106,7 @@ class Attribute {
 
   equals(targetAttribute) {
     if (!targetAttribute) return false;
-    const props = [ 'dataType', 'allowNull', 'defaultValue', 'primaryKey' ];
```
cyjake, line 109: `attribute.defaultValue` were left unset on purpose while migrating skylark models, because currently leoric will try to set attribute with the defaultValue when saving records. Theoretically, there is no need to set these values but since there might be existing code that relies on this behavior, let's keep it this way for now.

There might be models that have attribute.defaultValue set with values different than the actual default in database. When we're sure that there's none, we can add this check back.

test/unit/drivers/sqlite/pool.test.js
```diff
@@ -0,0 +1,17 @@
...
+    const pool = new Pool({
+      database: '/tmp/leoric.sqlite3',
+    });
+    let result;
+    pool.on('connection', function(connection) {
+      result = connection;
+    });
+    assert.equal(await pool.getConnection(), result);
+  });
+});
```
cyjake, line 17: emit `connection` event to enable users set session values before actually querying with the newly created connection, such as setting database session key.

src/drivers/sqlite/pool.js
```diff
@@ -0,0 +1,51 @@
...
+      if (connection.idle) {
+        connection.idle = false;
+        this.emit('acquire', connection);
+        return connection;
+      }
+    }
+    if (connections.length < options.connectionLimit) {
+      const connection = new Connection({ ...options, client, pool: this });
+      connections.push(connection);
+      this.emit('connection', connection);
```
JimmyDaddy, line 31: `connection` is not a verb, should we use `connect` or `alloc`?

src/drivers/abstract/attribute.js
JimmyDaddy: Agreed, we should avoid inserting values by default.

src/drivers/sqlite/pool.js
cyjake: The name was stolen from [mysqljs/mysql](https://github.com/mysqljs/mysql#connection), personally I'd agree with verb over noun, but let's try keep the event names consistent across drivers for now.

PR #173, opened 2021-09-13
refactor: entry.isFile() can only be called as method

cyjake:
att

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,10 +41,9 @@ async function findModels(dir) {
   const models = [];
 
   for (const entry of entries) {
-    const { name, isFile } = entry;
-    const extname = path.extname(name);
-    if (isFile() && ['.js', '.mjs'].includes(extname)) {
-      const model = require(path.join(dir, name));
+    const extname = path.extname(entry.name);
+    if (entry.isFile() && ['.js', '.mjs'].includes(extname)) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -41,10 +41,9 @@ async function findModels(dir) {
   const models = [];
 
   for (const entry of entries) {
-    const { name, isFile } = entry;
-    const extname = path.extname(name);
-    if (isFile() && ['.js', '.mjs'].includes(extname)) {
-      const model = require(path.join(dir, name));
+    const extname = path.extname(entry.name);
+    if (entry.isFile() && ['.js', '.mjs'].includes(extname)) {
+      const model = require(path.join(dir, entry.name));
```
cyjake, line 46: a bit weird that these change didn't trigger CI. Anyway, `entry.isFile()` requires this binding hence reverted. /cc @okbug

okbug: revert?

cyjake: It was changed in your previous pr https://github.com/cyjake/leoric/pull/170/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346

But `isFile()` method actually can not be called with out `entry`, hence I need to revert these change here.

xudafeng (approved):

PR #176, opened 2021-09-16
fix: set options.busyTimeout to mitigate SQLITE_BUSY

cyjake:
att


src/drivers/sqlite/pool.js
```diff
@@ -52,6 +55,12 @@ class Pool extends EventEmitter {
       task();
     }
   }
+
+  async end() {
+    for (const connection of this.connections) {
```
JimmyDaddy, line 60: `await Promise.allSettled(this.connections.map(async (connection) => await connection.end());` could be prefered.

cyjake: done

PR #178, opened 2021-09-22
refactor: declaration files (and tests); `connect(): Realm`

cyjake:
changed from `connect(): Bone` to `connect(): Realm` which makes more sense. This behavior is not actually used in any dependents hence is planed with a minor release


types/index.d.ts
```diff
@@ -121,15 +121,25 @@ interface WhereConditions {
...
-  columnType: string,
-  isNullable: boolean,
-  type: boolean | number | string | Date | JSON;
+  column: string;
+  columnType: string;
+  allowNull: boolean;
+  defaultValue: Literal;
+  primaryKey: boolean;
+  dataType: string;
+  jsType: boolean | number | string | Date | JSON;
```
JimmyDaddy, line 141: `Buffer` should be added

types/index.d.ts
```diff
@@ -543,6 +553,27 @@ interface ConnectOptions {
   models: string | Bone[];
 }
 
+interface InitOptions {
+  underscored?: boolean;
+  tableName?: string;
+  hooks?: Record<string, Function>;
```
JimmyDaddy, line 559: `hooks` should be `Record<string, AsyncFunction>`

types/index.d.ts
```diff
@@ -103,7 +103,7 @@ declare class Spell {
   toString(): string;
 }
 
-type Literal = null | boolean | number | string | Date
+type Literal = null | boolean | number | string | Date | JSON | ArrayBuffer;
```
cyjake, line 106: Buffer requires `@types/node`, added ArrayBuffer instead...

JimmyDaddy (approved):

types/index.d.ts
```diff
@@ -543,6 +553,31 @@ interface ConnectOptions {
...
 }
 
+interface InitOptions {
+  underscored?: boolean;
+  tableName?: string;
+  hooks?: {
+    [key in 'beforeCreate' | 'beforeBulkCreate' | 'beforeUpdate' | 'beforeSave' |  'beforeUpsert' | 'beforeRemove' ]: (options: QueryOptions) => Promise<void>
+  } | {
+    [key in 'afterCreate' | 'afterBulkCreate' | 'afterUpdate' | 'afterSave' | 'afterUpsert' | 'afterRemove' ]: (instance: Bone, result: Object) => Promise<void>
+  };
```
cyjake, line 563: added before/after names

It seems typescript removed `AsyncFunction`, added specific function type definitions. Static hooks not added yet.

PR #182, opened 2021-09-23
fix: selected instance isNewRecord should be false

JimmyDaddy:
attr


src/bone.js
```diff
@@ -1160,7 +1160,8 @@ class Bone {
         instance._getRawUnset().add(name);
       }
     }
-
+    // set isNewRecord to false
```
cyjake, line 1163: unnecessary comment, comment should be about the 'why' part

PR #183, opened 2021-09-23
fix: logic operate should adapt one argument

### LINKED ISSUE #181: [bug] $and AND $or with one value should work
```js

Model.find({
  where: {
     type: { $and: [ 'GOOD' ] }
  }
});

```
this will throw  `Cannot read property 'type' of undefined` at `src/drivers/abstract/spellbook.js:184:20`

JimmyDaddy:
fixes #181

### DIFF
--- a/test/unit/spell.test.js
+++ b/test/unit/spell.test.js
@@ -400,6 +400,13 @@ describe('=> Spell', function() {
     );
   });
 
+  it('where not in conditions', function() {
+    assert.equal(
+      Post.where('id != ?', [ 1, 2 ]).toString(),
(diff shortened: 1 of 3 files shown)


test/unit/spell.test.js
```diff
@@ -366,6 +400,13 @@ describe('=> Spell', function() {
     );
   });
 
+  it('where not in conditions', function() {
+    assert.equal(
+      Post.where('id != ?', [ 1, 2 ]).toString(),
+      'SELECT * FROM `articles` WHERE `id` NOT IN (1, 2) AND `gmt_deleted` IS NULL'
+    );
+  });
```
cyjake, line 408: image

PR #187, opened 2021-09-27
fix: sequelize mode support HAVING, and select fields raw sql support

JimmyDaddy:


src/spell.js
```diff
@@ -881,6 +883,7 @@ class Spell {
     const { columns, groups } = this;
 
     for (const name of names) {
+      console.log(name, 'name');
```
cyjake, line 886: console

test/unit/adapters/sequelize.test.js
```diff
@@ -454,6 +454,45 @@ describe('=> Sequelize adapter', () => {
...
+
+    result = await Post.findAll({
+      attributes: 'count(*) AS count',
+      group: [ 'title' ],
+      order: [[ 'title', 'desc' ]],
+    });
+    assert.deepEqual(result, [
+      { title: 'Tyrael', count: 1 },
+      { title: 'Leah', count: 2 },
+    ]);
```
cyjake, line 466: break into a new test case such as `it('Model.findAll({ group: [] })'`?

test/unit/adapters/sequelize.test.js
```diff
@@ -454,6 +454,45 @@ describe('=> Sequelize adapter', () => {
...
+    result = await Post.findAll({
+      attributes: 'count(*) AS count',
+      group: [ 'title' ],
+      order: [[ 'title', 'desc' ]],
+      having: raw('count(*) = 2')
+    });
+
+    assert.deepEqual(result, [
+      { title: 'Leah', count: 2 },
+    ]);
```
cyjake, line 495: same as above, I think breaking into separated test cases helps regression tests and debugging

cyjake: `having([ 'count = ?', 2 ])` should be supported as well, which can be achieved with `spell.$having(...having)`

src/spell.js
```diff
@@ -69,7 +70,7 @@ function parseObjectValue(value) {
   if (value && value.__raw) return { type: 'raw', value: value.value };
   // value maybe an object conditions
   if (isPlainObject(value)) {
-    const [ args ] = parseObjectConditions(value);
+    const args = parseObjectConditions(value);
     return args;
```
cyjake, line 74: `return parseObjectConditions(value);`

cyjake: this change breaks return type of parseObjectValue however.

src/spell.js
```diff
@@ -273,9 +274,11 @@ function parseObjectConditions(conditions) {
       });
     }
     else if (isLogicalCondition(value)) {
-      const condition = parseObjectValue(value);
-      normalizeLogicalCondition(condition, name);
-      result.push(condition);
+      const conds = parseObjectValue(value);
```
cyjake, line 277: maybe we can use `parseOperatorCondition(name, value)` here

cyjake: tried a bit, with no luck. I'd need more time to investigate on this.

JimmyDaddy: `parseOperatorCondition` seems can't parse logical operator and return object not object array

PR #191, opened 2021-09-29
refactor: object condition parsing and expression formatting

cyjake:
- object condition parsing => src/query_object.js
- expression formatting => src/expr_formatter.js

currently the `spell` object is still passed along the `formatXxx()` functions in expr_formatter.js. There might be two ways to refactor this:

- Turn `formatXxx()` into `ExprFormatter()` instances with `this.spell`;
- Turn expr objects into `Expr`, `OpExpr`, etc. with `this.spell`;

both requires lots of work but without little perf improvements, hence postponded further.

### DIFF
--- a/src/query_object.js
+++ b/src/query_object.js
@@ -1,0 +1,228 @@
+'use strict';
+
+const util = require('util');
+const { isPlainObject } = require('./utils');
+const { parseExpr } = require('./expr');
+// deferred to break cyclic dependencies
+let Spell;
+
+const OPERATOR_MAP = {
+  $between: 'between',
+  $eq: '=',
+  $gt: '>',
+  $gte: '>=',
+  $in: 'in',
+  $like: 'like',
+  $lt: '<',
+  $lte: '<=',
+  $ne: '!=',
+  $nin: 'not in',
+  $notBetween: 'not between',
+  $notLike: 'not like',
+  $notIn: 'not in'
+};
+
+/**
+ * Parse object values as literal or subquery
+ * @param {Object} value
+ * @returns {Array<Object>}
+ */
+function parseValue(value) {
+  if (value instanceof Spell) return { type: 'subquery', value };
+  if (value && value.__raw) return { type: 'raw', value: value.value };
+  return parseExpr('?', value);
+}
+
+/**
+ * Check if object condition is an operator condition, such as `{ $gte: 100, $lt: 200 }`.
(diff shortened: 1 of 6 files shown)


test/unit/spell.test.js
```diff
@@ -499,13 +499,6 @@ describe('=> Spell', function() {
     );
   });
 
-  it('orHaving with array attr', function() {
-    assert.equal(
-      Post.count().group('authorId').having([ 'count > ?', 10 ]).orHaving('count = 5').toString(),
```
cyjake, line 504: `having([])` is not supported in Spell, which is only supported in sequelize adapter with `having: []`

src/query_object.js
```diff
@@ -0,0 +1,228 @@
...
+      args.push(parseValue(val));
+    }
+
+    result.push({ type: 'op', name: operator, args });
+  }
+
+  return merge(result);
+}
+
+const LOGICAL_OPERATOR_MAP = {
```
JimmyDaddy, line 88: We should take those constants(`LOGICAL_OPERATOR_MAP`, `OPERATOR_MAP`, `AGGREGATOR_MAP`) in one place, such as `src/constants`? So that they can be used in other place too.

src/query_object.js
```diff
@@ -0,0 +1,228 @@
...
+ * // => { type: 'op', name: 'and', args: [ ... ]}
+ * @param {string} name
+ * @param {Object} condition
+ * @returns {Object}
+ */
+function parseOperator(name, condition) {
+  const result = [];
+
+  for (const $op in condition) {
+    const operator = $op === '$not' ? OPERATOR_MAP.$ne : OPERATOR_MAP[$op];
```
JimmyDaddy, line 69: It suddenly occurred to me that how can we process `NOT` or `!` operator? eg.
```SQL
SELECT * FROM table_name
WHERE NOT NOT (id > 100);
```

src/query_object.js
cyjake: I'd prefer to refactor it when necessary. Currently the constants can stay in this scope.

src/query_object.js
cyjake: fixed with https://github.com/cyjake/leoric/pull/191/commits/664316be961ac8981a8c2774ad97b4b2f2fc65b1

JimmyDaddy (approved):

PR #192, opened 2021-09-29
feat: support custom fields query and sequelize mode export rawAttributes

JimmyDaddy:


src/bone.js
```diff
@@ -1150,16 +1150,29 @@ class Bone {
...
 
+    const valueMap = {};
+
     for (const name in attributes) {
       const { columnName, jsType } = attributes[name];
       if (columnName in row) {
         // to make sure raw and rawSaved hold two different objects
         instance._setRaw(name, driver.cast(row[columnName], jsType));
         instance._setRawSaved(name, driver.cast(row[columnName], jsType));
+        valueMap[columnName] = columnName;
```
JimmyDaddy, line 1161: use key cache instead of `delete`

Issue #39, opened 2020-03-08
logger.logQuery should be guarded in case of error


cyjake:
instances like below:

```js
const realm = new Realm({
  logger: {
    logQuery() {
      throw new Error('asdf');
    }
  },
}
```

The error shall be logged (or output to console) and the query result shall return anyway

cyjake:
fixed by #222

PR #222, opened 2021-11-02
fix: logger.logQuery should be guarded in case of error

SmartOrange:
#39 logger.logQuery should be guarded in case of error

### DIFF
--- a/src/drivers/abstract/logger.js
+++ b/src/drivers/abstract/logger.js
@@ -12,7 +12,7 @@ class Logger {
   format(query, values, opts = {}) {
     const { command, sets } = opts;
 
-    if (['insert', 'upsert', 'update'].includes(command) && sets) {
+    if ([ 'insert', 'upsert', 'update' ].includes(command) && sets) {
       const { hideKeys } = this;
       const keys = Object.keys(sets);
       values = values.map((entry, i) => {
--- a/src/drivers/abstract/logger.js
+++ b/src/drivers/abstract/logger.js
@@ -13,7 +13,7 @@ class Logger {
   format(query, values, opts = {}) {
     const { command, sets } = opts;
 
-    if ([ 'insert', 'upsert', 'update' ].includes(command) && sets) {
+    if (['insert', 'upsert', 'update'].includes(command) && sets) {
       const { hideKeys } = this;
       const keys = Object.keys(sets);
       values = values.map((entry, i) => {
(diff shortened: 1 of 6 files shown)


test/unit/realm.test.js
```diff
@@ -106,6 +106,23 @@ describe('=> Realm', () => {
...
+        logger: {
+          logQuery(sql) {
+            throw new Error('asdf');
+          }
+        }
+      });
+      await realm.connect();
+      const res = await realm.driver.query('SELECT 1');
+      assert(res.rows);
+    });
```
cyjake, line 123: test/unit/drivers/abstract/index.test.js would be the better place to put this test case

src/drivers/abstract/logger.js
```diff
@@ -4,9 +4,10 @@ const debug = require('debug')('leoric');
 const SqlString = require('sqlstring');
 
 class Logger {
-  constructor(opts) {
+  constructor(opts = {}) {
     if (typeof opts === 'function') opts = { logQuery: opts };
-    Object.assign(this, { hideKeys: [] }, opts);
+    this._opts = opts;
```
cyjake, line 9: `this.options` is preferred.

src/drivers/abstract/logger.js
```diff
@@ -25,7 +26,15 @@ class Logger {
   }
 
   logQuery(sql, duration, opts) {
-    debug('[query] [%s] %s', duration, sql);
+    if (this._opts.logQuery) {
+      try {
+        this._opts.logQuery(sql, duration, opts);
+      } catch (error) {
+        console.error(error);
+      }
```
cyjake, line 34: The original implementation was intended to eliminate property checks as much as possible. Perhaps the new `logQuery` can be renamed as something like `tryLogQuery`?

cyjake: or just change the original function as the default `this.options.logQuery()`

test/unit/drivers/abstract/index.test.js
```diff
@@ -4,17 +4,31 @@ const assert = require('assert').strict;
...
+  it('should create logger by default', async function () {
     const driver = new AbstractDriver();
     assert.ok(driver.logger);
     assert.ok(driver.logger instanceof Logger);
   });
 
-  it('should accept custom logger', async function() {
-    class CustomLogger extends Logger {};
+  it('should accept custom logger', async function () {
+    class CustomLogger extends Logger { };
```
cyjake, line 15: a bit weird code format. can you persist the original format as eslint rule?

test/unit/realm.test.js
```diff
@@ -106,6 +106,7 @@ describe('=> Realm', () => {
       assert.equal(queries[0], 'SELECT 1');
     });
 
+
```
cyjake, line 109: one empty line is plenty hhhhh

cyjake (approved):

PR #229, opened 2021-11-13
fix: turn off subquery optimization if query criteria contains other column

### LINKED ISSUE #228: premature subquery optimization causes join query return insufficient data
Given data:

```js
await Post.bulkCreate([
  { id: 2, title: 'Archbishop Lazarus' },
  { id: 3, title: 'Archangel Tyrael' },
]);

await Comment.bulkCreate([
  { articleId: 2, content: 'foo' },
  { articleId: 2, content: 'bar' },
  { articleId: 3, content: 'baz' },
]);
```

Given query:

```js
await Post.include('comments').order('posts.id').where({ 
  'posts.title': { $like: 'Arch%' },
  'comments.content': 'baz',
});
```

Following SQL is generated:

```sql
 SELECT `posts`.*, `comments`.* FROM (SELECT * FROM `articles` WHERE `title` LIKE 'Arch%' AND `gmt_deleted` IS NULL LIMIT 1) AS `posts` LEFT JOIN `comments` AS `comments` ON `posts`.`id` = `comments`.`article_id` AND `comments`.`gmt_deleted` IS NULL WHERE `comments`.`content` = 'baz'
```

The subquery `SELECT * FROM `articles` WHERE `title` LIKE 'Arch%' AND `gmt_deleted` IS NULL LIMIT 1` rules out `Post  { id: 3, title: 'Archangel Tyrael' }` and keeps the wrong one `Post { id: 2, title: 'Archbishop Lazarus' }`

The subquery was intended to hoist the limit/offset to the main table to filter data as much as possible. It surely is a premature optimization that needs to be removed.

cyjake:
fixes #228

### DIFF
--- a/test/integration/suite/associations.test.js
+++ b/test/integration/suite/associations.test.js
@@ -36,10 +36,10 @@ describe('=> Associations', function() {
     stub = sinon.stub(logger, 'warn').callsFake((message) => {
       throw new Error(message);
     });
-    const posts = [
-      await Post.create({ title: 'Archbishop Lazarus' }),
-      await Post.create({ title: 'Leah' })
-    ];
+    const posts = await Post.bulkCreate([
+      { title: 'Archbishop Lazarus' },
+      { title: 'Leah' },
+    ]);
     const tags = await Promise.all(tagNames.map(name => Tag.create({ name, type: 0 })));
     const topics = await Promise.all(topicNames.map(name => Tag.create({ name, type: 1 })));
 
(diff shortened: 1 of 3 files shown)


test/integration/suite/associations.test.js
```diff
@@ -170,10 +170,26 @@ describe('scattered associations', function() {
...
+    assert.equal(posts[0].title, 'New Post 2');
+  });
+
+  it('should still limit the query if subquery limit is off', async function() {
+    const posts = await Post.include('comments').where({
+      'comments.content': { $like: '%oo%' },
+    }).limit(1);
+    // both posts have comments with content containing `oo` but only one should return
+    assert.equal(posts.length, 1);
+  });
```
cyjake, line 194: @JimmyDaddy test case to catch the forgotten offset/limit is now added

JimmyDaddy (approved):

PR #233, opened 2021-11-20
feat: make duration in precise milliseconds

fengmk2:
detail see eggjs/egg-logger#74

before:
```log
[query] [1] DROP TABLE IF EXISTS `clients`
```

after:
```log
[query] [1.478] DROP TABLE IF EXISTS `clients`
```


fengmk2:
image

微秒的耗时也可以以小数位展示出来

cyjake:
nice! 👍

PR #236, opened 2021-11-22
feat: make duration in precise milliseconds

fengmk2:
detail see eggjs/egg-logger#74

before:
```log
[query] [1] DROP TABLE IF EXISTS `clients`
```

after:
```log
[query] [1.478] DROP TABLE IF EXISTS `clients`
```


fengmk2:
pick from https://github.com/cyjake/leoric/pull/233

PR #245, opened 2021-12-21
fix: date format cast

JimmyDaddy:
```js

User {
  created_at: DATE
}

const u = User.create(...)
u.created_at = '2019-09-18 19:09:00.188' // should be u.created_at = '2019-09-18 19:09:00'

```

### DIFF
--- a/src/data_types.js
+++ b/src/data_types.js
@@ -220,7 +220,7 @@ class DATE extends DataType {
   constructor(precision, timezone = true) {
     super();
     this.dataType = 'datetime';
-    this.precision = precision;
+    this.precision = precision || 0;
     // PostgreSQL enables timestamp with or without time zone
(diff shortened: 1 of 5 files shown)


src/data_types.js
```diff
@@ -220,7 +220,7 @@ class DATE extends DataType {
   constructor(precision, timezone = true) {
     super();
     this.dataType = 'datetime';
-    this.precision = precision;
+    this.precision = precision || 0;
```
cyjake, line 223: precision should not default to 0. At least the timestamp precision is not default to 0 in neither mysql nor postgresql.

cyjake:
this pr should be superseded by https://github.com/cyjake/leoric/pull/246 , sorry for the late submit

PR #246, opened 2021-12-21
fix: information_schema.columns.datetime_precision

cyjake:
`new Model(values)` should have values rounded according to precision

### DIFF
--- a/src/data_types.js
+++ b/src/data_types.js
@@ -229,14 +229,24 @@ class DATE extends DataType {
   toSqlString() {
     const { precision } = this;
     const dataType = this.dataType.toUpperCase();
-    if (precision > 0) return `${dataType}(${precision})`;
+    if (precision != null && precision >= 0) return `${dataType}(${precision})`;
     return dataType;
+  }
+
+  _round(value) {
+    const { precision } = this;
+    if (precision != null && precision < 3 && value instanceof Date) {
+      const result = new Date(value);
+      result.setMilliseconds(result.getMilliseconds() % (10 ** precision));
+      return result;
+    }
+    return value;
   }
 
   cast(value) {
     if (value == null) return value;
-    if (value instanceof Date) return value;
-    return new Date(value);
+    if (!(value instanceof Date)) value = new Date(value);
+    return this._round(value);
   }
 
   uncast(value) {
@@ -258,13 +268,7 @@ class DATE extends DataType {
     if (!(value instanceof Date)) value = new Date(value);
     if (isNaN(value)) throw new Error(util.format('invalid date: %s', originValue));
 
-    const { precision } = this;
-    if (precision < 3) {
-      const result = new Date(value);
-      result.setMilliseconds(result.getMilliseconds() % (10 ** precision));
-      return result;
-    }
-    return value;
+    return this._round(value);
   }
 }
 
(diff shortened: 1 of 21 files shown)


src/raw.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+module.exports = class Raw {
+  constructor(value) {
+    this.value = value;
+    // consumed in expr_formatter.js
+    this.type = 'raw';
+  }
+};
```
cyjake, line 9: bit worried that the original raw data structure might be supplied by user input directly, such as `where({ name: { __raw: true, value: "'' OR 1" })`. Hence changed it to a specific js class that needs instantiation.

test/integration/suite/basics.test.js
```diff
@@ -172,20 +172,20 @@ describe('=> Basic', () => {
     });
 
     it('Bone.previousChanged(): raw VS rawPrevious', async function () {
-      const post = new Post({ title: 'Unknown', extra: 'hello' });
+      const post = new Post({ title: 'Unknown', extra: { greeting: 'hello' } });
```
cyjake, line 175: 'hello' won't be able to be converted into js object

src/drivers/postgres/schema.js
```diff
@@ -67,6 +69,7 @@ module.exports = {
         // https://www.postgresql.org/docs/9.5/infoschema-table-constraints.html
         primaryKey,
         unique: row.constraint_type === 'UNIQUE',
+        datetimePrecision: precision === 6 ? null : precision,
```
cyjake, line 72: postgres has datetime_precision default to 6 even if the precision is not specified in column definition. reset it back to `null` to work around existing test cases for now.

src/data_types.js
```diff
@@ -374,7 +385,12 @@ class JSON extends DataType {
     if (!value) return value;
     // type === JSONB
     if (typeof value === 'object') return value;
-    return global.JSON.parse(value);
+    try {
+      return global.JSON.parse(value);
+    } catch (err) {
+      console.error(new Error(`unable to cast ${value} to JSON`));
+      return value;
+    }
```
cyjake, line 393: JSON.cast will now fallback to original value

src/data_types.js
```diff
@@ -229,14 +229,24 @@ class DATE extends DataType {
...
-    if (precision > 0) return `${dataType}(${precision})`;
+    if (precision != null && precision >= 0) return `${dataType}(${precision})`;
     return dataType;
   }
 
+  _round(value) {
+    const { precision } = this;
+    if (precision != null && precision < 3 && value instanceof Date) {
+      const result = new Date(value);
+      result.setMilliseconds(result.getMilliseconds() % (10 ** precision));
```
JimmyDaddy, line 240: `result.getMilliseconds() % (10 ** precision)` seems not correct?
while `precision  = 2, Milliseconds = 864`, expect `86`
```js
864 % (10 ** 2) = 64
```
how about `Number(result.getMilliseconds()/ (10 ** (3- precision)).toFixed())`

cyjake: nice catch! updated with following solution:

```js
const divider = 10 ** (3 - precision);
 return new Date(Math.round(value.getTime() / divider) * divider);
```

JimmyDaddy (approved):

PR #251, opened 2021-12-31
fix: fix #250 format numeric result

### LINKED ISSUE #250: [bug] aggregator func returns string when use mysql2 as db driver
```js
const sum = await Book.sum('price');
typeof sum;// string when use mysql2
```

JimmyDaddy:


src/collection.js
```diff
@@ -88,7 +91,11 @@ function dispatch(spell, rows, fields) {
     if (type === 'alias' && args && args[0].type === 'func') {
       const row = rows[0];
       const result = row && (row[''] || row[table]);
-      return result && result[value] || 0;
+      // see https://www.w3schools.com/mysql/mysql_ref_functions.asp
+      if (AGGREGATORS.includes(args[0].name)) {
+        return Number(result && result[value]) || 0;
+      }
```
cyjake, line 97: should it fallback to `result[value]` instead of `0`?

cyjake:
v1.x is merged into master now. Consider pr to master instead?

PR #261, opened 2022-01-24
fix: fix #260 UPDATE with LIMIT and ORDER should be formatted(mysql only)

### LINKED ISSUE #260: [bug] UPDATE with LIMIT and ORDER should be formatted
```SQL
// mysql only
UPDATE users SET updated_at = "1977-01-01 19:22:22" ORDER BY id LIMIT 4
```

JimmyDaddy:
attr

### DIFF
--- a/src/drivers/abstract/spellbook.js
+++ b/src/drivers/abstract/spellbook.js
@@ -476,10 +476,20 @@ function formatUpdate(spell) {
     for (const condition of whereConditions) collectLiteral(spell, condition, values);
     chunks.push(`WHERE ${formatConditions(spell, whereConditions)}`);
   }
+
+  const extraOptions = this.formatUpdateExtraOptions(spell);
+  if (extraOptions.length) {
+    chunks.push(...extraOptions);
+  }
+
   return {
     sql: chunks.join(' '),
     values,
   };
+}
+
+function formatUpdateExtraOptions() {
+  return [];
 }
 
 /**
@@ -606,4 +616,6 @@ module.exports = {
   formatSelectWithoutJoin,
   formatUpdateOnDuplicate,
   formatReturning,
+  formatUpdateExtraOptions,
+  formatOrders
 };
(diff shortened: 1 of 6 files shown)


src/drivers/abstract/spellbook.js
```diff
@@ -476,12 +476,22 @@ function formatUpdate(spell) {
     for (const condition of whereConditions) collectLiteral(spell, condition, values);
     chunks.push(`WHERE ${formatConditions(spell, whereConditions)}`);
   }
+
+  const extraOptions = this.formatUpdateExtraOptions(spell);
+  if (extraOptions.length) {
+    chunks.push(...extraOptions);
+  }
```
cyjake, line 483: This bug fix shouldn't be sequelize adapter specific I think. Use cases like below should be supported as well:

```js
Post.update({ title: null }, { title: 'Untitled' }).order('id', 'desc').limit(10);
```

JimmyDaddy: It a MySQL only feature, this was supported at this PR, see 
* test/integration/suite/basics.test.js:685
* test/unit/spell.test.js:638

JimmyDaddy: > This bug fix shouldn't be sequelize adapter specific I think. Use cases like below should be supported as well:
> 
> ```js
> Post.update({ title: null }, { title: 'Untitled' }).order('id', 'desc').limit(10);
> ```
It's a MySQL only feature, this was supported at this PR, see

test/integration/suite/basics.test.js:685
test/unit/spell.test.js:638

test/integration/suite/basics.test.js
```diff
@@ -682,6 +682,21 @@ describe('=> Basic', () => {
       expect(posts.length).to.equal(affectedRows);
     });
 
+    it('Bone.update(where, values) should work with limit/offset and order', async function() {
+      if (Post.driver.type === 'mysql') {
```
JimmyDaddy, line 686: mysql only

cyjake (approved):

PR #262, opened 2022-01-24
fix: fixed #256 static create instance should check all default attri…

### LINKED ISSUE #256: [bug] static create instance should check all default attributes settings
```js
class User {
  name: { type: STRING, allowNull: false },
...
}

const u = await User.create({ email: 'ssss' });
/**
current throwing database error:

code: "ER_NO_DEFAULT_FOR_FIELD"
errno: 1364
sqlState: "HY000"
sqlMessage: "Field 'name' doesn't have a default value"

It should be:
throw new LeoricValidateError('notNull', 'name')
**/

```

JimmyDaddy:
…butes settings

### DIFF
--- a/src/adapters/sequelize.js
+++ b/src/adapters/sequelize.js
@@ -238,9 +238,6 @@ module.exports = Bone => {
     }
 
     static build(values, options = {}) {
-      if (options.validate !== false) {
-        this._validateAttributes(values);
-      }
       const { raw } = Object.assign({ raw: false, isNewRecord: true }, options);
       const { attributes } = this;
 
@@ -256,6 +253,7 @@ module.exports = Bone => {
       } else {
         instance = new this(values, options);
       }
+
       return instance;
     }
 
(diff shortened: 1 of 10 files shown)


src/adapters/sequelize.js
```diff
@@ -238,9 +238,6 @@ module.exports = Bone => {
     }
 
     static build(values, options = {}) {
```
JimmyDaddy, line 240: sequelize Model.build doesn't validate attributes, only validate values before persisting data

src/bone.js
```diff
@@ -736,7 +724,12 @@ class Bone {
     }
 
     if (opts.validate !== false) {
-      this._validateAttributes();
+      const validateValues = {};
+      for (const key in attributes) {
+        if (attributes[key].primaryKey) continue;
+        validateValues[key] = this[key];
+      }
+      this._validateAttributes(validateValues);
```
JimmyDaddy, line 732: It should validate all set values before create an instance

cyjake (approved):

PR #271, opened 2022-02-14
fix: attribute.defaultValue should be set when init attributes

cyjake:
when loading attributes from information_schema, the `attribute.defaultValue` should be set as well to be compliant with the validation change introduced by https://github.com/cyjake/leoric/pull/262/files

### DIFF
--- a/test/dumpfile.sql
+++ b/test/dumpfile.sql
@@ -9,7 +9,7 @@ CREATE TABLE `articles` (
   `extra` text,
   `thumb` varchar(1000) DEFAULT NULL,
   `author_id` bigint(20) DEFAULT NULL,
-  `is_private` tinyint(1) DEFAULT 0,
+  `is_private` tinyint(1) DEFAULT 0 NOT NULL,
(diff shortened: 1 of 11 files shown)


test/dumpfile.sql
```diff
@@ -9,7 +9,7 @@ CREATE TABLE `articles` (
   `extra` text,
   `thumb` varchar(1000) DEFAULT NULL,
   `author_id` bigint(20) DEFAULT NULL,
-  `is_private` tinyint(1) DEFAULT 0,
+  `is_private` tinyint(1) DEFAULT 0 NOT NULL,
```
cyjake, line 12: v2.0.2 and v2.0.3 will throw LeoricValidationError in such column settings, because the attribute.defaultValue is not present, but the values were validated before create.

JimmyDaddy (approved):

cyjake:
the line that misses test coverage is a debug output at rare cases in sqlite driver, ignored for now

PR #276, opened 2022-02-19
fix: previousChanges should check instance is new record or not while specific attributes' values were undefined

JimmyDaddy:
```js
// title: Untitled, authorId: undefined
const post = new Post({ title: 'Untitled' });
assert.deepEqual(post.previousChanges('title'), {});
assert.deepEqual(post.previousChanges('authorId'), {});
post.title = 'MHW';
// authorId was undefined
post.authorId = 100;
await post.save();
// post is not a new record now
assert.deepEqual(post.previousChanges('authorId'), { authorId: [ null, 100 ] });

```

### DIFF
--- a/src/bone.js
+++ b/src/bone.js
@@ -183,7 +183,6 @@ class Bone {
     }
 
     if (args.length > 1) {
-      // execute validators
       this.#raw[name] = value instanceof Raw ? value : attribute.cast(value);
       this.#rawUnset.delete(name);
       return this;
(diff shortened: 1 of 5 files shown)


src/bone.js
```diff
@@ -371,15 +371,15 @@ class Bone {
    */
   previousChanges(name) {
     if (name != null) {
-      if (this.#rawUnset.has(name) || this.#rawPrevious[name] === undefined || !this.hasAttribute(name)) return {};
+      if (this.#rawUnset.has(name) || (this.#rawPrevious[name] === undefined && this.isNewRecord) || !this.hasAttribute(name)) return {};
```
cyjake, line 374: It looks like `this.isNewRecord` covers `this.#rawPrevious[name] === undefined` already, can we give it a try?

btw, this line is a bit too much now, I believe a multiple line if block is necessary once the column width exceeds some threshold like 90 or 120

PR #278, opened 2022-02-20
feat: add types for realm

luckydrq:
- Make the last two parameters optional of `Realm.prototype.define()`.
- Add `sync` method signature.

### DIFF
--- a/test/start.sh
+++ b/test/start.sh
@@ -29,7 +29,7 @@ function integration {
 ##
 # definition type tests
 function dts {
-  npx tsc
+  npx tsc --esModuleInterop
   run "$(ls test/types/*.test.js)";
 }
 
--- a/test/start.sh
+++ b/test/start.sh
@@ -29,7 +29,7 @@ function integration {
 ##
 # definition type tests
 function dts {
-  npx tsc --esModuleInterop
+  npx tsc
   run "$(ls test/types/*.test.js)";
 }
 
--- a/test/types/realm.test.ts
+++ b/test/types/realm.test.ts
@@ -36,6 +36,36 @@ describe('=> Realm (TypeScript)', function () {
   });
 
   describe('realm.sync(options)', async function() {
+    it('options should be optional', async function() {
+      assert.doesNotThrow(async () => {
+        const { STRING } = realm.DataTypes;
+        realm.define('User', { name: STRING });
+        await realm.sync();
+      });
+    });
 
+    it('`force` can be passed individually', async function() {
+      assert.doesNotThrow(async () => {
+        const { STRING } = realm.DataTypes;
+        realm.define('User', { name: STRING });
+        await realm.sync({ force: true });
+      });
+    });
+
+    it('`alter` can be passed individually', async function() {
+      assert.doesNotThrow(async () => {
+        const { STRING } = realm.DataTypes;
+        realm.define('User', { name: STRING });
+        await realm.sync({ alter: true });
+      });
+    });
+
+    it('`force` and `alter` can be passed together', async function() {
+      assert.doesNotThrow(async () => {
+        const { STRING } = realm.DataTypes;
+        realm.define('User', { name: STRING });
+        await realm.sync({ force: true, alter: true });
+      });
+    });
   });
 });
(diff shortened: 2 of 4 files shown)


types/index.d.ts
```diff
@@ -634,8 +639,8 @@ export default class Realm {
   define(
     name: string,
     attributes: Record<string, AttributeMeta>,
-    options: InitOptions,
-    descriptors: Record<string, Function>,
+    options?: InitOptions,
+    descriptors?: Record<string, Function>,
```
cyjake, line 643: descriptors aren't `Record<string, Function>`, the type of it is more like:

```ts
interface Descriptor {
  get() {},
  set() {},
  // not sure of the syntax above
}
Record<string, Descriptor>
```

It should be clear once you include some unit tests in test/types

cyjake: following syntax is possible as well, don't know how to express the type of it though:

```ts
Foo.init({}, {}, {
  get foo() {},
  set foo() {},
});
```

luckydrq: > descriptors aren't `Record<string, Function>`, the type of it is more like:
> 
> ```ts
> interface Descriptor {
>   get() {},
>   set() {},
>   // not sure of the syntax above
> }
> Record<string, Descriptor>
> ```
> 
> It should be clear once you include some unit tests in test/types

This `descriptors` param seems to get passed in [getOwnPropertyDescriptors()](https://github.com/cyjake/leoric/blob/master/src/bone.js#L1553) call. It is a plain Object according to the [spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors).

types/index.d.ts
```diff
@@ -593,7 +593,7 @@ export class Bone {
   toObject(): InstanceValues<this>;
 }
 
-interface ConnectOptions {
+export interface ConnectOptions {
```
luckydrq, line 596: need to port it to egg-orm

types/index.d.ts
cyjake: yep, the descriptors are more like `{ get foo() {} }` rather than full format property descriptors (which makes the parameter name misleading). 

Anyway, `Record<string, Function>` couldn't be right, take the `{ get foo() {} }` for example, the type of this object isn't `Record<string, Function>`, rather than something like `{ get [key: string](): any, set [key: string](): any }` (not sure if this works).

We can have the correct type definition of descriptor parameter once the test cases are ready.

luckydrq: > yep, the descriptors are more like `{ get foo() {} }` rather than full format property descriptors (which makes the parameter name misleading).
> 
> Anyway, `Record<string, Function>` couldn't be right, take the `{ get foo() {} }` for example, the type of this object isn't `Record<string, Function>`, rather than something like `{ get [key: string](): any, set [key: string](): any }` (not sure if this works).
> 
> We can have the correct type definition of descriptor parameter once the test cases are ready.

Seems `getter/setter` are [not fully supported in TypeScript](https://github.com/microsoft/TypeScript/issues/3745#issuecomment-118924225) yet.

cyjake: yep, and looks like tsc is happy with `Record<string, Function>`. Let's focus on the rest of the changes, make sure these changes have test covered as well.

test/types/realm.test.ts
```diff
@@ -0,0 +1,41 @@
...
+          this.attribute('name', value);
+        }
+      });
+    });
+  });
+
+  describe('realm.sync(options)', async function() {
+
+  });
+});
```
cyjake, line 41: @luckydrq you can re-validate the typescript test quickly with `npm run test:dts`

luckydrq: @cyjake Done. Please check again.

test/start.sh
```diff
@@ -29,7 +29,7 @@ function integration {
 ##
 # definition type tests
 function dts {
-  npx tsc
+  npx tsc --esModuleInterop
```
cyjake, line 32: compiler options need to be put in ./tsconfig.json in project root

PR #280, opened 2022-02-23
fix: add missing `password` field for `ConnectOptions`

luckydrq:
- Add missing `password` for `ConnectOptions`.
- Fix the return type of `Realm.prototype.define()`.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -599,6 +599,7 @@ export interface ConnectOptions {
   host?: string;
   port?: number | string;
   user?: string;
+  password?: string;
   database: string;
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -644,7 +645,7 @@ export default class Realm {
     attributes: Record<string, DataTypes<DataType> | AttributeMeta>,
     options?: InitOptions,
     descriptors?: Record<string, Function>,
-  ): Bone;
+  ): typeof Bone;
```
luckydrq, line 648: image

PR #283, opened 2022-02-25
fix: instance beforeUpdate hooks should not modify any Raw if there are no Raw  assignment  in them

JimmyDaddy:
att


src/setup_hooks.js
```diff
@@ -129,9 +129,28 @@ function addHook(target, hookName, func) {
...
+              changeRaw[name] = this.attribute(name);
+            }
+          }
+          await func.apply(this, args);
+          // revert instance after before hooks
+          Object.keys(originalRaw).forEach((key) => {
+            const current = this.attribute(key);
+            // raw[key] may changed in beforeUpdate hooks
+            if (current !== originalRaw[key] && current !== changeRaw[key]) return;
+            this.attribute(key, originalRaw[key]);
```
JimmyDaddy, line 149: skip custom setter because it may throw error or execute biz logic

cyjake (approved):

PR #296, opened 2022-03-22
fix: drop column if not defined in attributes when alter table

### LINKED ISSUE #294: app.model.sync方法不能正确的同步model到数据库
### 复现步骤
通过`javascript model`来更新数据库表结构。
1. 创建普通的egg项目。并且配置数据库链接。使其能够正常链接数据库
2. 创建一个 `user.js`的数据模型
 ```javascript
module.exports = function(app, Model) {
    const { STRING } = Model.DataTypes;
    class User extends Model {
        static attributes = {
            name: STRING,
            password: STRING,
            avatar: STRING,
        }
    }
    return User
  }
```
4. 在egg app.js生命周期中执行model同步数据库
 ```javascript
  async serverDidReady() {
        console.log('同步表结构中...')
        await this.app.model.sync({ alter: true})        
        console.log('表结构同步完成...')
    }
```
5. 数据库表创建数据成功
6. 修改`user.js`为如下
  ```javascript
 module.exports = function(app, Model) {
    const { STRING } = Model.DataTypes;
    class User extends Model {
        static attributes = {
            name: STRING,
            password: STRING,
            // avatar: STRING, // 屏蔽掉avatar属性
        }
    }
    return User
  }
 ```
7. 重新运行egg项目
### 实际结果
数据`user`表中任然存在 `avatar`列。
###  期望结果
数据库表中不存在 `avatar`列表

### 问题溯源
`leoric/src/bone.js` 1020行。执行`compare`比对有问题。
```javascript
Object.defineProperties(this, looseReadonly({
      timestamps,
      table,
      primaryKey,
      columns,
      attributeMap,
      associations,
      tableAlias,
      synchronized: Object.keys(compare(attributes, columnMap)).length === 0,// 此处对比有问题
    }));
```

cyjake:
also (partially) added column comment support to close #294

### DIFF
--- a/src/bone.js
+++ b/src/bone.js
@@ -37,16 +37,24 @@ function looseReadonly(props) {
 
 function compare(attributes, columnMap) {
   const diff = {};
+  const columnNames = new Set();
 
   for (const name in attributes) {
     const attribute = attributes[name];
     const { columnName } = attribute;
+    columnNames.add(columnName);
 
     if (!attribute.equals(columnMap[columnName])) {
       diff[name] = {
         modify: columnMap.hasOwnProperty(columnName),
         ...attribute,
       };
+    }
+  }
+
+  for (const columnName in columnMap) {
+    if (!columnNames.has(columnName)) {
+      diff[columnName] = { remove: true };
     }
   }
 
@@ -254,8 +262,8 @@ class Bone {
 
   /**
    * get actual update/insert columns to avoid empty insert or update
-   * @param {Object} data 
-   * @returns 
+   * @param {Object} data
+   * @returns
    */
   static _getColumns(data) {
     if (!Object.keys(data).length) return data;
--- a/src/data_types.js
+++ b/src/data_types.js
@@ -17,17 +17,20 @@ class DataType {
     const {
       STRING, TEXT,
       DATE, DATEONLY,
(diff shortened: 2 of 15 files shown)


src/bone.js
```diff
@@ -50,6 +52,12 @@ function compare(attributes, columnMap) {
     }
   }
 
+  for (const columnName in columnMap) {
+    if (!columnNames.has(columnName)) {
+      diff[columnName] = { remove: true };
```
barretlee, line 57: `removed`?

src/data_types.js
```diff
@@ -17,17 +17,21 @@ class DataType {
...
+    const params = [];
+    for (let i = 0; i < matches.length; i++) {
+      if (matches[i] != null) params[i] = parseInt(matches[i], 10);
+    }
 
     switch (dataType) {
       case 'varchar':
       case 'char':
-        return new STRING(length);
+        console.log(params);
```
barretlee, line 33: remove console.log

src/bone.js
cyjake: there is `{ modify: true }` already, so... on the other hand, the properties were added in a way that is more like to tell driver what to update not what happened in model attributes.

src/data_types.js
cyjake: nice catch

JimmyDaddy (approved):

PR #306, opened 2022-04-22
feat: export more ts type definitions and  use deep-equal module

JimmyDaddy:
att


types/index.d.ts
```diff
@@ -187,7 +196,49 @@ declare class Pool {
...
+  format(spell: Spell<typeof Bone, number| ResultSet | null>): Object;
+
+  formatInsert(spell: Spell<typeof Bone, number>): Object;
+  formatSelect(spell: Spell<typeof Bone, ResultSet | null>): Object;
+  formatUpdate(spell: Spell<typeof Bone, number>): Object;
+  formatDelete(spell: Spell<typeof Bone, number>): Object;
+  formatUpsert(spell: Spell<typeof Bone, null>): Object;
+}
+
+declare class AbstructDriver {
```
cyjake, line 236: AbstractDriver

Issue #317, opened 2022-06-23
[bug] init models with bone class should work


JimmyDaddy:
```js
// under app/packageA
const { Bone } = require('leoric');
class User extends Bone {};

// under app/packageB
const realm = new Realm({
   ...options,
   models: [  User ]
});

await realm.connect(); // throw error driver is undefined

```

PR #318, opened 2022-06-24
fix: fix [bug] init models with bone class should work #317

JimmyDaddy:
init models in different node modules context


test/unit/realm.test.js
```diff
@@ -1055,6 +1055,41 @@ describe('=> Realm', () => {
...
+        port: process.env.MYSQL_PORT,
+        user: 'root',
+        database: 'leoric',
+        models: [ Post, User ],
+      });
+      await realm.connect();
+      assert(!!Post.driver);
+      assert(!!Post.options);
+      assert(!!User.driver);
+      assert(!!User.options);
```
cyjake, line 1091: `assert()` takes truthy values, no need to coerce them to boolean

src/realm.js
```diff
@@ -75,6 +75,10 @@ async function loadModels(Spine, models, opts) {
   const schemaInfo = await Spine.driver.querySchemaInfo(database, tables);
 
   for (const model of models) {
+    // assign driver if model's driver not exist
+    if (!model.driver) model.driver = Spine.driver;
+    // assign options if model's options not exist
+    if (!model.options) model.options = Spine.options;
```
cyjake, line 81: was `Spine.driver` initialized elsewhere already?

JimmyDaddy: `Spine.driver` initialized in `new Realm(options)`,  but `model.driver/options` may not be initialized,  because `model` may not extended from the `Spine` aforementioned, this is an edge case.

PR #319, opened 2022-06-28
feat: change DataTypes to ts and complete decorators type definitions

JimmyDaddy:
att


src/decorators.ts
```diff
@@ -53,8 +62,28 @@ export function Column(options: ColumnOption | DataTypes<DataType> = {}) {
...
+    if(!Object.keys(customDescriptor).length) return;
+    
+    customDescriptor.enumerable = true;
+    customDescriptor.configurable = true;
+
+    const defaultDescriptor = Object.getOwnPropertyDescriptor(target, propertyKey);
+    Object.defineProperty(target, propertyKey, {
+      ...defaultDescriptor,
+      ...customDescriptor,
+    })
```
cyjake, line 86: are we going to support custom `getter`/`setter`s with `{ getter() {}, setter() {} }`? I think we've already got `{ get() {}, set() {} }` covered.

cyjake (approved):

PR #323, opened 2022-07-15
test: switch the auth protocol of test mysql database 

cyjake:
... to the one that's supported


cyjake:
```js
Error: ER_NOT_SUPPORTED_AUTH_MODE: Client does not support authentication protocol requested by server; consider upgrading MySQL client
```

node-mysql doesn't support the default auth protocol in mysql v8.x

JimmyDaddy (approved):

PR #324, opened 2022-07-20
feat: add leoric_bone meta data to Bone

JimmyDaddy:
https://github.com/eggjs/egg-orm/blob/7086680f533b9efa70db55e06345adbb5ac321f5/lib/loader.js#L46
```
if (!(OriginModelClass && OriginModelClass.prototype && OriginModelClass.prototype instanceof Model)) {
  return OriginModelClass;
}
// => 
if (!(Realm.isLeoricBone(OriginModelClass)) {
  return OriginModelClass;
}
```
model class could be multiple inheritance, `OriginModelClass.prototype instanceof Model` cannot be a proper way to validate a model is a leoric bone

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,7 @@ const Realm = require('./src/realm');
 const Decorators = require('./src/decorators');
 const Raw = require('./src/raw');
 const { MysqlDriver, PostgresDriver, SqliteDriver, AbstractDriver } = require('./src/drivers');
+const { isBone } = require('./src/utils');
 
 /**
  * @typedef {Object} RawSql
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -14,6 +14,7 @@ const Realm = require('./src/realm');
 const Decorators = require('./src/decorators');
 const Raw = require('./src/raw');
 const { MysqlDriver, PostgresDriver, SqliteDriver, AbstractDriver } = require('./src/drivers');
+const { isLeoricBone } = require('./src/utils');
```
cyjake, line 17: can we change this to `isBone()`? The `Leoric` part in the function name seems unnecessary.

cyjake: the rest looks good to me

cyjake (approved):

PR #329, opened 2022-08-24
fix: projects might have strictPropertyInitialization set to true

cyjake:
att

### DIFF
--- a/src/data_types.ts
+++ b/src/data_types.ts
@@ -25,17 +25,8 @@ export interface AbstractDataType<T> {
  */
 
 export abstract class DataType {
-  dataType: string;
+  dataType: string = '';
   dataLength?: string | number;
-
-  /**
-   * Check if params is instance of DataType or not
-   * @param {*} params
-   * @returns {boolean}
-   */
-  static is(params: any): boolean {
-    return params instanceof DataType;
-  }
 
   /**
    * cast raw data returned from data packet into js type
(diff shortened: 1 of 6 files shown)


src/data_types.ts
```diff
@@ -25,18 +25,9 @@ export interface AbstractDataType<T> {
  */
 
 export abstract class DataType {
-  dataType: string;
+  dataType: string = '';
```
cyjake, line 28: `compilerOptions.strictPropertyInitialization` complains about unset class field

cyjake: which is quite common in our test models, such as 

```ts
class Foo extends Bone {
  @Column()
  id: bigint;
}
```

added a default value to make downstream projects happy for now

src/data_types.ts
```diff
@@ -598,8 +575,23 @@ export class DataTypes extends DataType {
...
+    }); 
+  }
+
+  /**
+   * Check if params is instance of DataType or not
+   * @param {*} params
+   * @returns {boolean}
+   */
+   static is(params: any): boolean {
+    return params instanceof DataType;
```
cyjake, line 594: further refactor the default export `class DataTypes` and the base `class DataType`, which can be imported as `import DataTypes, { DataType } from './data_types';`

existing code that relies on `const { DataTypes } = require('./data_types')` is changed to `const { default: DataTypes } = require('./data_types')` accordingly

JimmyDaddy (approved):

PR #335, opened 2022-08-25
fix: should skip loading models that is loaded before

cyjake:
att


src/realm.js
```diff
@@ -83,7 +83,6 @@ async function loadModels(Spine, models, opts) {
     const columns = schemaInfo[model.physicTable] || schemaInfo[model.table];
     if (!model.attributes) initAttributes(model, columns);
     model.load(columns);
-    Spine.models[model.name] = model;
```
cyjake, line 86: moved to L160 to make sure all of the models passed in are registered onto `Bone.models`

JimmyDaddy (approved):

PR #337, opened 2022-08-27
feat: refactor type definitions to export SequelizeBone, complete spell type definitions and fix index hints logic

JimmyDaddy:
* refactor type definitions to export `SequelizeBone` for sequelize adaptor
* complete spell  and hints definitions, fix hints logic
* virtual column should not be added to unset in case of custom getter cannot be serialize
```ts
class User {
 @Column()
 id: bigint;
 
 // custom will be ignored when toObject()/toJSON() execute, so make it a VIRTUAL column
 get custom(): string {
    return 'v';
 }
}
```


src/bone.js
```diff
@@ -1291,8 +1291,8 @@ class Bone {
     }
 
     for (const name in attributes) {
-      const { columnName } = attributes[name];
-      if (!(columnName in row)) {
+      const attribute = attributes[name];
+      if (!(attribute.columnName in row) && !attribute.virtual) {
```
JimmyDaddy, line 1295: virtual column should not be added to unset in case of custom getter cannot be serialize
```js
class User {
 @Column()
 id: bigint;
 
 // custom will be ignored when toObject()/toJSON() execute, so make it a VIRTUAL column
 get custom(): string {
    return 'v';
 }
}
```

src/decorators.ts
```diff
@@ -37,7 +37,7 @@ function findType(tsType) {
   }
 }
 
-export function Column(options?: ColumnOption | DATA_TYPE<DataType>) {
+export function Column(options?: ColumnOption | DATA_TYPE<DataType> | DataType) {
```
JimmyDaddy, line 40:
```ts
class User {
 // types invokable should work
 @Column(DataTypes.STRING(12))
 id: bigint;
}
```

src/types/bone.d.ts
```diff
@@ -0,0 +1,410 @@
...
+interface SyncOptions {
+  force?: boolean;
+  alter?: boolean;
+}
+
+interface TransactionOptions {
+  connection: Connection;
+}
+
+export class AbstractBone {
```
JimmyDaddy, line 21: `static find` `static findOne` `static update` and others overloaded methods  cannot be override by subclass definitions because of different args required, so create a `AbstractBone` as superclass of `Bone` and `SequelizeBone`

src/types/bone.d.ts
```diff
@@ -0,0 +1,410 @@
...
+   * @example
+   * Bone.get(8)
+   * Bone.find({ foo: { $gt: 1 } }).get(42)
+   */
+  static get<T extends typeof AbstractBone>(this: T, index: number): Spell<T, InstanceType<T> | null>;
+
+  /**
+   * SELECT rows ORDER BY id ASC LIMIT 1
+   */
+  static first: Spell<typeof AbstractBone, AbstractBone | null>;
```
JimmyDaddy, line 140: Didn't get a proper way to fix this type issue(same as `await Bone.all/last`、 `Bone.unscoped` and `SequelizeBone.Model`  )
```ts
class Post extends Bone {
   @Column()
   title: string;
}

const p = await Post.first;
// error here, title not a property of post
p.title;
```

JimmyDaddy: `await Post.find().first` can work temporarily because of `Spell.first/last`

cyjake: yes, static getter in TypeScript seems can't have polymorphic this. https://github.com/microsoft/TypeScript/issues/50301

src/types/bone.d.ts
cyjake: 👌

should we put this dts file along side with src/bone.js like the others (src/spell.d.ts, src/adapters/sequelize.d.ts) as well?

JimmyDaddy: done

cyjake (approved):

PR #345, opened 2022-09-19
fix: Literal type should not contain `object` & ResultSet should be derived from instance values

cyjake:
use Record<string, any> instead

### DIFF
--- a/src/adapters/sequelize.d.ts
+++ b/src/adapters/sequelize.d.ts
@@ -128,8 +128,8 @@ export class SequelizeBone extends AbstractBone {
    */
   static bulkBuild<T extends typeof SequelizeBone>(this:T, valueSets: Array<Values<T>>, options?: BoneOptions): Array<InstanceType<T>>;
 
-  static count<T extends typeof SequelizeBone>(this: T, name?: string): Spell<T, ResultSet | number>;
-  static count<T extends typeof SequelizeBone>(this: T, conditions?: SequelizeConditions<T>): Spell<T, ResultSet | number>;
+  static count<T extends typeof SequelizeBone>(this: T, name?: string): Spell<T, ResultSet<T> | number>;
+  static count<T extends typeof SequelizeBone>(this: T, conditions?: SequelizeConditions<T>): Spell<T, ResultSet<T> | number>;
 
   static decrement<T extends typeof SequelizeBone>(
     this: T,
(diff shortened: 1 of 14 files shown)


src/adapters/sequelize.d.ts
```diff
@@ -128,8 +128,8 @@ export class SequelizeBone extends AbstractBone {
    */
   static bulkBuild<T extends typeof SequelizeBone>(this:T, valueSets: Array<Values<T>>, options?: BoneOptions): Array<InstanceType<T>>;
 
-  static count<T extends typeof SequelizeBone>(this: T, name?: string): Spell<T, ResultSet | number>;
-  static count<T extends typeof SequelizeBone>(this: T, conditions?: SequelizeConditions<T>): Spell<T, ResultSet | number>;
+  static count<T extends typeof SequelizeBone>(this: T, name?: string): Spell<T, ResultSet<T> | number>;
+  static count<T extends typeof SequelizeBone>(this: T, conditions?: SequelizeConditions<T>): Spell<T, ResultSet<T> | number>;
```
cyjake, line 132: to be able to prompt group fields in scenarios like:

```ts
const results = await Post.group('title').count();
results[0].title // should be able to tell the type of `title`
```

src/types/common.d.ts
```diff
@@ -3,7 +3,7 @@ import { CommonHintsArgs } from '../hint';
 import { AbstractDataType, DataType } from '../data_types';
 import { AbstractBone } from './abstract_bone';
 
-export type Literal = null | undefined | boolean | number | bigint | string | Date | object | ArrayBuffer;
+export type Literal = null | undefined | boolean | number | bigint | string | Date | Record<string, any> | ArrayBuffer;
```
cyjake, line 6: bigint types gets checked correctly since `object` type is removed from Literal union type

src/types/common.d.ts
```diff
@@ -168,13 +166,10 @@ export type WhereConditions<T extends typeof AbstractBone> = {
...
 }
 
-export type Values<T extends typeof AbstractBone> = {
-  [Property in keyof Extract<InstanceType<T>, Literal>]?: Literal;
-}
+// https://stackoverflow.com/a/68077021/179691
+export type PickTypeKeys<Obj, Type, T extends keyof Obj = keyof Obj> = ({ [P in keyof Obj]: Obj[P] extends Type ? P : never })[T];
 
-export type InstanceValues<T> = {
-  [Property in keyof Extract<T, Literal>]?: Extract<T, Literal>[Property]
```
cyjake, line 176: `Extract<T, Literal>` extracts type Literal from union T, which does not match what we need here.

JimmyDaddy (approved):

PR #347, opened 2022-09-21
fix: deletedAt should always be checked when associating models

cyjake:
att

### DIFF
--- a/src/bone.js
+++ b/src/bone.js
@@ -1016,6 +1016,9 @@ class Bone {
       if (columnInfo && attribute.type instanceof DataTypes.DATE && attribute.type.precision == null) {
         attribute.type.precision = columnInfo.datetimePrecision;
       }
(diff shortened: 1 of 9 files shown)


src/bone.js
```diff
@@ -1016,6 +1016,9 @@ class Bone {
       if (columnInfo && attribute.type instanceof DataTypes.DATE && attribute.type.precision == null) {
         attribute.type.precision = columnInfo.datetimePrecision;
       }
+      // if (columnInfo && columnInfo.defaultValue != null && attribute.defaultValue === undefined) {
+      //   attribute.defaultValue = columnInfo.defaultValue;
+      // }
```
cyjake, line 1021: loading `defaultValue` from database, not activated yet due to not tested thoroughly, pending on @JimmyDaddy

JimmyDaddy: hmm....I'll check it

chenhui5416 (approved):
good

PR #353, opened 2022-09-27
fix: this.attribute(name) should fallback to Literal

cyjake:
```ts
class Foo {
  @Column(TEXT)
  get settings(): Record<string, any> {
    // this.attribute('settings') should return the original union type here
    const text = this.attribute('settings') as string;
    if (!text) return null;
    try {
      return JSON.parse(text);
    } catch {
      return null;
    }
  }
}
```

### DIFF
--- a/src/types/abstract_bone.d.ts
+++ b/src/types/abstract_bone.d.ts
@@ -241,15 +241,15 @@ export class AbstractBone {
    * bone.attribute('foo');     // => 1
    * bone.attribute('foo', 2);  // => bone
    */
-  attribute<T, Key extends keyof T>(this: T, name: Key, value: T[Key]): void;
-  attribute<T, Key extends keyof T>(this: T, name: Key): T[Key];
+  attribute<T, Key extends keyof T>(this: T, name: Key, value: Literal): void;
+  attribute<T, Key extends keyof T, U extends T[Key]>(this: T, name: Key): U extends Literal ? U : Literal;
 
   /**
    * Get the original attribute value.
    * @example
    * bone.attributeWas('foo')  // => 1
    */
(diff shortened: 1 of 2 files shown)


src/types/abstract_bone.d.ts
```diff
@@ -241,15 +241,15 @@ export class AbstractBone {
    * bone.attribute('foo');     // => 1
    * bone.attribute('foo', 2);  // => bone
    */
-  attribute<T, Key extends keyof T>(this: T, name: Key, value: T[Key]): void;
-  attribute<T, Key extends keyof T>(this: T, name: Key): T[Key];
+  attribute<T, Key extends keyof T>(this: T, name: Key, value: Literal): void;
+  attribute<T, Key extends keyof T, U extends T[Key]>(this: T, name: Key): U extends Literal ? U : Literal;
```
cyjake, line 245: If U is assignable to Literal, return U, else return Literal. This shall prevent this.attribute() calls in getter gives impractical type such as `object | Function`.

test/types/basics.test.ts
```diff
@@ -48,7 +48,20 @@ describe('=> Basics (TypeScript)', function() {
     summary: string;
 
     @Column(TEXT)
-    settings: string;
+    get settings(): Record<string, any> {
+      const text = this.attribute('settings') as string;
```
cyjake, line 52: technically Record<string, any> is still assignable to LIteral but somehow this works, this.attribute('settings') returns Literal here.

cyjake: might because of the extra `U extends T[Key]` that derives a new type from `T[Key]`?

JimmyDaddy: it seems `text`'s type is `Literal | this["settings"]`

JimmyDaddy (approved):

PR #364, opened 2022-10-28
fix: `tinytext` type error

JimmyDaddy:
and type definitions


src/decorators.ts
```diff
@@ -54,7 +54,7 @@ export function Column(options: ColumnOption | DATA_TYPE<DataType> | DataType =
 
     // target refers to model prototype, an internal instance of `Bone {}`
     const model = target.constructor as any;
-    if (!Object.hasOwnProperty('attributes') || !model.attributes) {
```
cyjake, line 57: 😅

cyjake (approved):

PR #368, opened 2022-11-08
fix: `cast/uncast` STRING and TEXT with non-string type value, and some type definitions

JimmyDaddy:
att

### DIFF
--- a/src/data_types.ts
+++ b/src/data_types.ts
@@ -37,7 +37,7 @@ export abstract class DataType {
   /**
    * uncast js value into database type with precision
    */
-  uncast(value: any, _strict?: boolean): any {
+  uncast(value: any): any {
     return value;
   }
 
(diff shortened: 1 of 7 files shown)


src/data_types.ts
```diff
@@ -66,6 +66,16 @@ class STRING extends DataType {
     return chunks.join(' ');
   }
 
+  cast(value: any): any {
```
JimmyDaddy, line 69:
```ts
class Note extends Bone {
  static attributes = {
    meta: TEXT,
  }
}

// fixed:
// logger output: INSERT `notes` (`meta`) VALUES (`name`='bloodborne', `type`='Cthulhu');
// actual SQL: INSERT `notes` (`meta`) VALUES ('[object object]');
const meta = { name: 'bloodborne', type: 'Cthulhu' };
const note = await Note.create({ meta });
```

cyjake (approved):

src/data_types.ts
cyjake: 👍

PR #374, opened 2022-12-26
fix: adapt facebook hermes or metro bundler for loop

JimmyDaddy:
and  type definitions

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -98,16 +98,18 @@ function dispatch(spell, rows, fields) {
 
   const results = new Collection();
   for (const row of rows) {
-    const result = {};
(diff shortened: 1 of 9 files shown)


src/collection.js
```diff
@@ -98,16 +98,18 @@ function dispatch(spell, rows, fields) {
 
   const results = new Collection();
   for (const row of rows) {
-    const result = {};
```
JimmyDaddy, line 101: facebook hermes or metro bundler compiles the previous  version code to 
截屏2022-12-26 13 54 43

it lose the original code's logic

cyjake (approved):

PR #382, opened 2023-03-24
fix: select everything when forming subspell

cyjake:
otherwise the columns might be quolified unnecessarily, or worse, insufficient for further joining


test/integration/suite/sharding.test.ts
cyjake, line 1: 这个看上去是 false alarm，先回滚掉了

JimmyDaddy (approved):

PR #383, opened 2023-04-20
fix: order clause need be repeated on the main query

cyjake:
to retain the expected result order

### DIFF
--- a/src/spell.js
+++ b/src/spell.js
@@ -389,10 +389,9 @@ class Spell {
     Object.assign(this, {
       whereConditions: [],
       groups: [],
-      orders: [],
+      orders: JSON.parse(JSON.stringify(this.orders)),
       havingConditions: [],
       joins: {},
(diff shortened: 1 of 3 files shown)


JimmyDaddy (approved):

src/spell.js
```diff
@@ -389,10 +389,9 @@ class Spell {
     Object.assign(this, {
       whereConditions: [],
       groups: [],
-      orders: [],
+      orders: JSON.parse(JSON.stringify(this.orders)),
```
fengmk2, line 392: haha, deep copy in the faster way.

cyjake: 是的…… 看了一下 structuredClone 要 v17 才有

fengmk2: https://github.com/cyjake/leoric/pull/384 try to use structuredClone first.

PR #384, opened 2023-04-23
feat: try to use structuredClone first

fengmk2:


src/utils/index.js
```diff
@@ -50,11 +50,19 @@ function isBone(bone) {
   return metaValue === true;
 }
 
+function deepClone(value) {
+  if (typeof structuredClone === 'function') {
+    return structuredClone(value);
+  }
+  return JSON.parse(JSON.stringify(value));
+}
```
cyjake, line 58:
```js
const deepClone = typeof structuredClone === 'function' 
  ? structuredClone 
  : (value) => JSON.parse(JSON.stringify(value))
```
节省一点运行时判断 :-D

fengmk2: 有道理

cyjake (approved):

PR #385, opened 2023-04-25
fix: avoid ts type detect error

fengmk2:
> Operator '>' cannot be applied to types 'string | number' and 'number'.


fengmk2:
https://github.com/cnpm/cnpmcore/pull/425 复现例子

cyjake (approved):

cyjake:
我发个版本

PR #391, opened 2023-06-08
fix: wrong export from sequelize

vagusX:


src/browser.ts
```diff
@@ -48,7 +47,7 @@ export { Collection };
 export { DataTypes };
 export { Logger };
 export { Spell };
-export { sequelize };
+export * from './adapters/sequelize';
```
cyjake, line 50: adapters/sequelize 只有 module.exports = function sequelize() {} 诶，export * from 这种写法拿到的输出会是啥

vagusX: image

主要是 `SequelizeBone`

JimmyDaddy (approved):

src/browser.ts
cyjake: 👌

PR #414, opened 2023-12-07
feat: support JSON_VALUE(j, '$.a' RETURNING CHAR(50))

cyjake:
It looks like the performance of querying properties in native JSON data type in MySQL can be boosted with specific index, as long as the index is created with JSON_VALUE(field, path RETURNING type) that casts the property value to certain type.

To enable this optimization, the expression parser and formatter in Leoric needs to be updated, hence proposed this pr.

- https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-21.html#mysqld-8-0-21-json
- https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#function_json-value

### DIFF
--- a/src/expr.js
+++ b/src/expr.js
@@ -96,6 +96,20 @@ const PRECEDENCES = [
   ['or']
 ];
 
+const RETURNING_TYPES = [
+  'float',
+  'double',
+  'decimal',
+  'signed',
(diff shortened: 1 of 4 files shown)


JimmyDaddy (approved):
LGTM

PR #433, opened 2025-02-08
fix: object results should have attribute values casted as well

cyjake:
```typescript
class Post {
  @Column(DataTypes.JSONB)
  extra: Record<string, unknown>
}

const [post] = await Post.select('extra', "JSON_EXTRACT(extra, '$.foo') AS foo");
console.log(typeof post.extra); // expected "object", actually got "string"

### DIFF
--- a/src/bone.js
+++ b/src/bone.js
@@ -1194,7 +1194,11 @@ class Bone {
     for (const key in data) {
       const value = data[key];
       const attribute = attributeMap[key];
-      result[attribute ? attribute.name : key] = value;
+      if (attribute) {
+        result[attribute.name] = attribute.cast(value);
+      } else {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/bone.js [hunk 1]
```diff
@@ -1194,7 +1194,11 @@ class Bone {
     for (const key in data) {
       const value = data[key];
       const attribute = attributeMap[key];
-      result[attribute ? attribute.name : key] = value;
+      if (attribute) {
+        result[attribute.name] = attribute.cast(value);
```
cyjake, line 1198: 也可能导致预期之外的转换结果，发版之前得在应用里预先检查一遍

chenhui5416 (approved):

PR #463, opened 2026-04-01
fix: guard undefined columns in loadModels and remove unsafe database cast

cyjake:
- throw clear error when schema info is missing for a model table
- use this.options.database instead of unsafe `as string` cast on opts.database
- suppress TS7016 for JS-only test model import
- add test for bigint primary key dedup in joined results

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,4 @@ docs/Gemfile.lock
 package-lock.json
 .DS_Store
 .repo-dev
-
+.vscode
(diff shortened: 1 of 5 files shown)


cyjake:
@copilot apply changes based on the comments in [this thread](https://github.com/cyjake/leoric/pull/463#pullrequestreview-4041494446)

JimmyDaddy (approved):
LGTM

PR #466, opened 2026-07-18
Benchmark Proxy class field handling

cyjake:
## Summary

- Add a focused benchmark for the Proxy used to preserve ES2022 class field behavior.
- Cover construction, row instantiation, native class field initializers, hot getter/setter access, and `Object.defineProperty()` paths.
- Add `npm run benchmark` and `npm run benchmark:proxy` scripts.

## Benchmark Findings

Latest local run:

```text
Node v22.21.1, 100,000 iterations, 7 measured runs, 3 warmup runs
GC before samples: yes

case                                     direct ops/s    proxy ops/s   direct ns/op    proxy ns/op   proxy/direct
-----------------------------------------------------------------------------------------------------------------
construct with values                         765,763        753,488          1,306          1,327          1.02x
instantiate row                               300,535        232,531          3,327          4,300          1.29x
construct + native undefined fields           700,901        575,798          1,427          1,737          1.22x
getter hot path                            14,354,755      3,521,881             70            284          4.08x
setter hot path                            10,144,303      2,721,422             99            367          3.73x
defineProperty attribute initializer        8,347,535      4,444,642            120            225          1.88x
defineProperty non-attribute               13,043,762      5,238,391             77            191          2.49x
```

The main regression risk appears to be steady-state attribute access on Proxy instances rather than plain construction. Query row hydration is moderately slower, while hot getter/setter loops are several times slower in this isolated benchmark.

## Verification

- `npm run benchmark:proxy`
- `./node_modules/.bin/eslint benchmarks/proxy-class-fields.js`

### DIFF
--- a/benchmarks/proxy-class-fields.js
+++ b/benchmarks/proxy-class-fields.js
@@ -1,0 +1,287 @@
+'use strict';
+
+const { performance } = require('perf_hooks');
+const { Bone, DataTypes } = require('../lib');
+const AbstractDriver = require('../lib/drivers/abstract').default;
+const { hasLoadedAttributesKey } = require('../lib/abstract_bone');
+
+const DEFAULT_ITERATIONS = 100000;
+const DEFAULT_RUNS = 7;
+const DEFAULT_WARMUP_RUNS = 3;
+
+const iterations = readPositiveInt('ITERATIONS', DEFAULT_ITERATIONS);
+const runs = readPositiveInt('RUNS', DEFAULT_RUNS);
+const warmupRuns = readPositiveInt('WARMUP_RUNS', DEFAULT_WARMUP_RUNS);
+
+Bone.driver = new AbstractDriver();
+Bone.options = { define: {} };
+
+const ATTRIBUTES = {
+  name: DataTypes.STRING,
+  title: DataTypes.STRING,
+  email: DataTypes.STRING,
+  status: DataTypes.INTEGER,
+  level: DataTypes.INTEGER,
+  views: DataTypes.INTEGER,
+  score: DataTypes.INTEGER,
+  bio: DataTypes.TEXT,
+};
+
+const COLUMNS = [
+  { columnName: 'id', columnType: 'bigint', dataType: 'bigint', isNullable: 'NO' },
+  { columnName: 'name', columnType: 'varchar(255)', dataType: 'varchar', isNullable: 'YES' },
+  { columnName: 'title', columnType: 'varchar(255)', dataType: 'varchar', isNullable: 'YES' },
+  { columnName: 'email', columnType: 'varchar(255)', dataType: 'varchar', isNullable: 'YES' },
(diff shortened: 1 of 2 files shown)


cyjake:
## Follow-up: design implications

After looking beyond the benchmark itself, I do not think a long-lived instance `Proxy` should become Leoric's default architecture. It solves the hardest compatibility case transparently, but it charges every model operation for a collision that only occurs during derived-class field initialization.

One important correction to the apparent escape hatch: using TypeScript `declare` avoids emitting runtime fields, so it avoids the correctness conflict, but it does **not currently avoid the Proxy cost**. `loadAttribute()` sets `hasLoadedAttributesKey`, and the constructor returns a Proxy for every loaded model once that marker is set, regardless of whether any runtime class field is present.

The ecosystem does not appear to have a better transparent interception trick:

- Sequelize still documents `declare` as required because public class fields shadow its generated accessors.
- TypeORM maps ordinary instance properties and lets the ORM control construction/hydration; entity constructor arguments must be optional.
- Prisma avoids active-record class instances and returns plain objects.

The practical direction for Leoric is therefore to move the work out of the instance hot path.

### Suggested incremental design

1. Make `declare` / no-runtime-field mode genuinely Proxy-free and keep it as the fast TypeScript default.
2. Add one internal construct-then-hydrate primitive and route ORM-created instances through it (`instantiate`, `create`, bulk creation, validation instances, and Sequelize-compatible `build`).
3. After construction, inspect only mapped attributes for own data properties, capture any initializer values, delete those properties so the prototype accessors are visible again, then apply caller/database values last.
4. Offer an explicit final-class wrapper such as `Bone.model(Model)` or a class decorator for users who need direct `new Model(values)` with vanilla runtime fields.
5. Retain the current instance Proxy behavior as an explicit compatibility mode for fully transparent direct construction.

The central primitive would conceptually be:

```js
const instance = Reflect.construct(Model, []);
finalizeClassFields(instance); // capture defaults and remove attribute shadows
setValues(instance, values);   // supplied values win over field defaults
return instance;               // ordinary object, no long-lived Proxy
```

This works especially well for query hydration because `instantiate()` already constructs an empty instance and fills its raw values afterward.

There is a real language boundary here: a base constructor cannot run code after all derived class fields have initialized. Without a Proxy, unmodified direct `new User(values)` cannot simultaneously preserve prototype attribute accessors and transparently support vanilla fields. A factory or final-class wrapper is needed for that case.

Longer term, a major version could consider making ordinary own properties authoritative and moving casting/validation to hydration and persistence boundaries. Leoric already determines changes by comparing current values with saved snapshots, so that direction may be feasible, but it is a larger behavioral redesign.

My recommendation for the near term is: **Proxy-free default, construct-time repair for ORM-controlled paths, explicit wrapper/decorator for direct construction, and instance Proxy only as an opt-in compatibility mode.**

JimmyDaddy (approved):

PR #467, opened 2026-07-18
Replace instance Proxy with model compilation

cyjake:
## Summary

- remove the long-lived instance `Proxy` and the per-construction class-field repair wrapper
- keep `extends Bone` plus `declare` as the preferred direct TypeScript path, including ordinary TypeScript test fixtures
- compile `@Model()` and `realm.define(Class, attributes)` definitions once into fresh runtime models
- add own readiness markers and a prescriptive `LeoricClassFieldError` for unsafe direct ES2022 fields
- preserve `realm.define("Name", attributes)` as the generated-model path
- document the complete design in `docs/design/es2022-class-fields.md`

## Model paths

1. `realm.define("User", attributes)` generates a normal realm `Bone` subclass.
2. A registered `class User extends Bone` with `declare` fields is used directly with no replacement.
3. `@Model()` and `realm.define(Class, attributes)` call the shared one-time model compiler. The definition constructor and instance fields never execute, so mapped own properties cannot shadow Leoric accessors.

Direct models are marked ready by realm registration, `connect()`, or explicit attribute loading. Every concrete leaf needs its own marker. The first ORM-created direct instance is checked for mapped own fields; successful validation is cached per model.

Compiled definitions copy supported prototype descriptors, public static and symbol-backed configuration, and model metadata. Custom constructors, private instance fields, and ordinary instance initializers are intentionally outside the initial compiled-class contract.

## Performance

`RUNS=9 COMPILATION_ITERATIONS=20000 npm run benchmark:model-compilation` on Node.js 22.21.1, Apple Silicon, with 100,000 steady-state iterations:

| One-time operation | Direct | `@Model()` compiled | Realm compiled |
| --- | ---: | ---: | ---: |
| define/compile fresh class | 605 ns (1.00x) | 2,670 ns (4.41x) | 2,775 ns (4.59x) |

| Case | Generated | `@Model()` compiled | Realm compiled |
| --- | ---: | ---: | ---: |
| construct with values | 1.00x | 1.03x | 1.01x |
| instantiate row | 1.00x | 1.01x | 1.01x |
| getter hot path | 0.97x | 0.89x | 0.90x |
| setter hot path | 0.99x | 0.94x | 0.93x |

Compilation adds about 2.1 microseconds once per model. Compiled construction remained within 3% of direct construction, hydration within 1%, and getter/setter performance showed no regression. `declare` remains the recommendation because it avoids compilation and preserves constructor and inheritance identity. `benchmark:proxy` remains available for comparison.

## Validation

- rebased onto latest `master` (`0326d92`)
- `npm exec tsc -- --noEmit`
- `npm run lint -- --quiet`
- focused class-field suites: 25 passing; `src/model.ts` at 100% statements/branches/functions/lines
- TypeScript fixture compiled and executed with ES2022 `useDefineForClassFields: true`
- same fixture compiled and executed with legacy `useDefineForClassFields: false`
- native JavaScript fields evaluated directly on Node.js 18, 20, 22, and 24
- converted direct-model TypeScript suites: 67 passing
- SQL.js integration suite with raw JavaScript fixtures: 161 passing
- Sequelize non-database coverage: 16 passing; 6 MySQL-dependent hooks unavailable locally
- `npm run build:types:ts4.9`
- `npm run prepack:browser`
- `npm pack --dry-run`
- `npm run benchmark:model-compilation`
- Codecov project and patch checks pass

MySQL and PostgreSQL services are unavailable in the local environment; their dependent suites were not completed locally. The full database-backed CI matrix passes.


src/model.ts
```diff
@@ -0,0 +1,86 @@
...
+  });
+  Object.defineProperty(CompiledModel, COMPILED_MODEL, {
+    value: true,
+    configurable: false,
+    enumerable: false,
+    writable: false,
+  });
+
+  markModelClassReady(CompiledModel);
+  markModelClassFieldsChecked(CompiledModel);
```
JimmyDaddy: [P2] Do not pre-mark a compiled child as fields-checked when its ready base is still unchecked

`registerModel()` and the `models` option mark a direct model ready before its first mapped-own-field scan. If a compiled child extends such a parent, this loop stops at the ready parent and the resulting class still executes the parent's native field initializers, but this unconditional marker causes `assertModelClassFields()` to skip validation forever.

I reproduced this with a ready-but-unchecked parent containing native field `name`: `Child.instantiate({ name: 'Ada' })` produced `{ ownName: true, propertyName: undefined, rawName: 'Ada' }`. Subsequent property writes bypass Leoric's accessor and diverge from raw storage.

Please only pre-mark the compiled model when the selected base has already passed its own field check, or leave the child unchecked so its first ORM-managed instance is scanned. A regression test covering a registered unchecked parent plus compiled child would catch this.

src/abstract_bone.ts
```diff
@@ -420,6 +418,7 @@ export class AbstractBone {
     opts = opts ?? {};
     const data = Object.assign({}, values);
     const instance = new this(data);
+    assertModelClassFields(instance);
```
JimmyDaddy, line 421: [P2] Apply the class-field guard to instance persistence paths as well

This covers static `Model.create()`, but the documented `new Model(values).save()` and direct `instance.create()` paths never call `assertModelClassFields()` (`_save()` / `_create()` / instance update/upsert have the same gap).

I reproduced an unsafe native direct model where `new UnsafeDirect({ name: 'Grace' }).create()` returned without an error while the instance was `{ ownName: true, propertyName: undefined, rawName: 'Grace' }`. A later `instance.name = 'Lin'` updates only the own property, so persistence can still use the stale raw value.

Please guard the instance persistence boundaries too, and add a native ES2022 field regression test for `new Model(...).save()` / `instance.create()`.

JimmyDaddy (commented):
Found two reproducible data-consistency gaps in the new class-field guard. Details are attached inline.

JimmyDaddy (approved):

